PR #117, opened 2015-03-09
[needs discussion] Historical divisions

handlers:
The script at `scripts/country-us/historical-districts.py` grabs historical districts from the [congress-legislators repository](https://github.com/unitedstates/congress-legislators) and adds OCD ids for these districts to `identifiers/country-us/historical/unitedstates_legislators_autogenerated/historical_congressional_districts.csv`.

Two things:

1. I have spot checked the missing districts and they seem sane. Would love to get more eyes on that.
2. Adding several historical districts to a state that currently has a single at-large rep messes with the naming on that state's at-large district. Happy to try to patch `compile.py` to handle this. See the diff on Montana as an example:
```
-ocd-division/country:us/state:mt/cd:1,Montana,ocd-division/country:us/state:mt,Montana's at-large congressional district,,,,,,,,
+ocd-division/country:us/state:mt/cd:1,Montana's 1st congressional district,ocd-division/country:us/state:mt,Montana's at-large congressional district,,,,,,,,
```


paultag:
So, the end result is looking good -- Let's sync on that at large thing, I think I'm missing something

We should fix the indentation on that script, though -- let's run flake8 over it, and fix up the spacing. I see some places indenteded with two, other places with 3 - let's clean that up to be 4 everywhere per tab in

handlers:
OK, cool, I will clean this up — thanks for the heads w/r/t flake8 and spacing.

handlers:
@paultag this is now largely flake8 compliant (except a couple of crazy long strings — let me know if there's a conventional way to make said strings flake8 compliant). The latest commits also address the at-large naming issue.

scripts/country-us/historical-districts.py
```diff
@@ -0,0 +1,193 @@
...
+            district_dict[state].append(district[-1])
+        else:
+            district_dict[state] = [district[-1]]
+    return [st for st in district_dict if len(district_dict[st]) == 1]
+
+
+def write_missing_districts(missing_districts):
+    new_file_path = os.path.join(
+        SCRIPT_DIRECTORY,
+        '../../identifiers/country-us/historical/unitedstates_legislators_autogenerated/historical_congressional_districts.csv')
```
jamesturk: traditional way to make this PEP8 compliant is

    new_file_path = os.path.join(SCRIPT_DIRECTORY, 
                                             '../../identifiers/country-us/historical/'
                                             '/unitedstates_legislators_autogenerated/historical_congressional_districts.csv')

(this got formatted a bit wrong, but the idea is to split the string and use implicit concatenation)

handlers:
Flake8 has no problems after 657bc4f

paultag:
nice. Thanks, Sam!

paultag:
Actually, hum! When I re-run this script, it looks to churn the file,
is there some non-determinism in the ordering? Could we get this to sort
consistantly so we don't rewrite the entire file when we run it?

Otherwise looks flawless.

handlers:
Good call, @paultag. We're still rewriting the file when we run it, but it now consistently outputs the same thing.

paultag:
@handlers nice. Perfect, thanks!

PR #121, opened 2015-04-24
Standardize NH floterial district naming, and assign them GeoIDs

mileswwatkins:
New Hampshire floterial SLDLs should have the same naming scheme in the `whitelist` as in the `census_autogenerated`.


mileswwatkins:
@paultag: Validation check added. Pull 'er in?

@jpmckinney: I'm not sure which fields ought to be unique in Canada, but you can add those to the `COUNTRY_UNIQUE_FIELDS` dictionary.

paultag:
Nice. Thanks, @mileswwatkins

PR #124, opened 2015-08-26
[AU] Add territory to Fraser federal electorate

chrismytton:
I'm not sure if this is the correct territory for Fraser or if there was a reason it didn't have a territory. But this seems to be consistent with the rest of the data in this file.


dracos:
Fraser is in two territories, Australian Capital Territory and the Jervis Bay Territory, which is presumably why the ID doesn't list one.

jpmckinney:
Is the Fraser in each territory the *same* division or are they formally/bureaucratically different? For example, in Canada we have towns that cross provincial boundaries, but they are the same town. If the two Frasers are the same, the preferred ID should omit a territory, but new IDs for each territory's Fraser can be minted with a sameAs property pointing to the preferred ID that omits the territory.

dracos:
Sorry, I guess I wasn't clear. There is only one Fraser: https://en.wikipedia.org/wiki/Division_of_Fraser . The Australian Capital Territory has two representatives, Canberra and Fraser. Jervis Bay Territory is the port for Canberra, and is covered by Fraser for representation purposes.

jpmckinney:
https://en.wikipedia.org/wiki/Jervis_Bay_Territory

> it is counted as part of the ACT for the purpose of the ACT's representation in the Senate; and it forms part of the Division of Fraser for House of Representatives purposes.

Sounds good then.

PR #126, opened 2015-09-22
Add Chicago precincts

evz:
Just noticed that these were not in here so, there ya go.


jpmckinney:
How are these sourced? Also, can you run `python3 scripts/compile.py us` to update the primary country-us.csv?

evz:
They are extracted from the shapefile on the Chicago dataportal: https://data.cityofchicago.org/Facilities-Geographic-Boundaries/Precincts-current-/uvpq-qeeq

I'll run the compile step

jpmckinney:
If you can resolve conflicts, I'll review.

PR #131, opened 2016-10-24
Update country-us.csv with new NC OCD IDs.

ccongchen:
Thanks for the quick approval, @jamesturk! I am under the impression that the individual OCD ID file changes need to be separated from changes to the country-us.csv (or maybe just a separate commit). Let know what is the best way to do this and thanks for your patience. :)


ccongchen:
Great. Thank you so much @jamesturk !

PR #137, opened 2016-11-22
Added Dakota, Orleans, and Philippine US territories

showerst:
Added these 3, with start/end dates from Wikipedia. 

First PR here, running 
`compile.py us `
went cleanly, let me know if I need to commit the generated csv, or did anything wrong.

### DIFF
--- a/identifiers/country-us/us_territories.csv
+++ b/identifiers/country-us/us_territories.csv
@@ -1,9 +1,11 @@
-id,name,sameAs,sameAsNote,validThrough,validFrom
-ocd-division/country:us/territory:as,American Samoa
-ocd-division/country:us/territory:gu,Guam
-ocd-division/country:us/territory:mp,Northern Mariana Islands
-ocd-division/country:us/territory:pr,Puerto Rico
-ocd-division/country:us/territory:vi,US Virgin Islands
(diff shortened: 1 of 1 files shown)


jpmckinney:
Can you:

* add sameAs `ocd-division/country:us/territory:la` to Orleans
* rename to "Orleans Territory"
* generate and commit country-us.csv
* remove empty sameAsNote column

showerst:
When I make the sameAs change and try to regenerate, I get:

> ERROR: ocd-division/country:us/territory:ot is sameAs ocd-division/country:us/territory:la which does not exist 

Do I do the correct one with no sameAs, then add a second line for the incorrect one with a sameAs ?

showerst:
I went ahead and did that, my diff on countries-us.csv is massive though, let me know if that's abnormal.

jpmckinney:
Oops, I meant `ocd-division/country:us/state:la`

Also, do we need `dk`? Does anyone use that OCDID?

showerst:
dk comes from congress-legislators as well. (they use OL, DK, PI). After seeing your edit with LA I understand what you were getting at w/ the alias now, i'll remove the DK one and recommit.

I see why my git diff is so massive, creating an entry with validFrom made it add a new field to every row in the csv.

Thanks for walking me through all of these steps.

PR #146, opened 2017-06-21
+833 new divisions across 30 states

rsimoes:
This is a small subset of the divisions created during CTCL's 2016 Ballot Information Project. It contains no new district types. There was also a fair amount of diff noise in identifiers/country-us.csv from running compile.py, which caused line terminators to change from LF to CRLF.


jpmckinney:
@jamesturk Who did we decide governs the US division IDs?

jamesturk:
I think tentatively we'd said that @djbridges was working to figure that out among interested parties, I believe @rsimoes you work w/ Donny right?  I just scanned this and it looks superficially OK

lbirdpew:
Hi,

Is there an ETA on this being approved?  Just curious

jamesturk:
@rsimoes any word on a governance process for these?  I don't want to hold these up, they look reasonable and I know you all are using them in production but wasn't sure if we're close to a review process for these or not

PR #147, opened 2017-06-23
Accept http redirects and fix nl scraper

cmyr:
@jpmckinney I... kind of don't know what I'm doing here, but I made the various expections go away? 😕 I'm not sure if you would prefer just changing the statcan urls to http instead of adding a fairly silly helper function.

I didn't commit the CSVs because this is a speculative PR. The diffs on the CSVs are basically just a whole bunch of deletions.

Please advise?


jpmckinney:
Instead of adding that code, why not replace the URLs? Sometimes URLs break by redirecting to a special 404 page that returns HTTP 200, so we don't want the code to add another layer to discovering that issue.

cmyr:
@jpmckinney my thinking was that the redirects might be some temporary certificate issue which would revert at some point in the near future, but I don't think my solution is particularly elegant or anything. Happy to defer to your suggestion. 👍

jpmckinney:
I was being optimistic with StatCan. I changed the protocol to `https` and it worked for a time, but their own links never used `https`.

cmyr:
@jpmckinney fair enough. I changed the URLs back. I have a local error around encodings with the `pdftotext` call but I checked the actual resource fetching so I suspect that's something related to my sort of hacky (macOS) install of `xpdf`.

jpmckinney:
Ah, I think you need `pdftotext` from `poppler`, not from `xpdf` (if you're using Homebrew).

cmyr:
@jpmckinney Okay, poppler works but I'm getting an exception around a missing identifier. Is there some identifier dataset that needs updating?

jpmckinney:
I'd have to see the error.

cmyr:
```

**Comment by cmyr** [comment]
@jpmckinney 
errror:

```
./ca_municipal_subdivisions.rb:753:in `block in census_subdivisions_sk': sk:T:GRAND~COULEE (RuntimeError)
	from ./ca_municipal_subdivisions.rb:708:in `each'
	from ./ca_municipal_subdivisions.rb:708:in `census_subdivisions_sk'
	from ./ca_municipal_subdivisions.rb:360:in `has_children'
	from /Users/cmyr/Dev/work/opennorth/ocd-division-ids/scripts/country-ca/utils.rb:145:in `run'
	from ./ca_municipal_subdivisions.rb:761:in `<main>'
```

`identifier` is an empty string.

jpmckinney:
`sk:T:GRAND~COULEE` is the relevant part of that. It's notifying you that it can't disambiguate a Census subdivision in Saskatchewan (`sk`) that is a town (`T`) and whose name is normalized to `GRAND~COULEE`. If I look in ca_census_subdivisions.csv, it looks the issue is that Grand Coulee is a village (`VL`) based on the Census, not a town as in the PDF downloaded from gov.sk.ca earlier in the method. I had to add a few lines to account for that exception: https://github.com/opencivicdata/ocd-division-ids/commit/975f9a5cb2b698072e86776c1117b3f3618316d2#diff-30f34a3cdf7c19605452392187a34885L692

cmyr:
Okay, sounds good. The csv's are canonical? I was under the impression that we were _generating_ the csvs here, am I mistaken? I'll admit I'm not totally clear on what's going on in this corner of the project.

jpmckinney:
All the CSVs are generated, but there is hierarchy / dependency between some. The Census (as a product of the government) has higher trust than other sources, so we defer to it always.

The maintenance of some of these scripts is tricky as automatically reconciling diverse sources isn't easy.

cmyr:
@jpmckinney okay, thanks for the clarification.

With your changes, this runs without errors. The output is a few small changes, and then about 1000 deletions from `identifiers/country-ca/ca_municipal_subdivisions-posts_count.csv`. Does this seem reasonable to you? Should I add the modified `csv`s to this PR?

jpmckinney:
Ah, yeah, the code that contributed those deleted lines for Quebec (I think) can no longer reach the data source. Unless there were other changes to the file, just revert the `posts_count` file. Please add the CSVs to the PR, thanks.

cmyr:
@jpmckinney okay, this should be good to go afaik.

PR #150, opened 2017-10-31
ca: Update census datasource

cmyr:
@jpmckinney Still pretty 🤷‍♂️ with this project but this looks about right?


cmyr:
@jpmckinney also just requested access to [this gdoc](https://docs.google.com/spreadsheets/d/1ihCIDc9EtvxF7kzPg3Yk6e928DN7JzaycH92IBYr0QU/edit#gid=25) from an email address "colin*@gmail.com".

jpmckinney:
Looks like `ca_federal_electoral_districts-2013.csv` removed duplicates (the duplicates probably appeared because the old shapefile had multiple features for the same riding) and made two typo corrections (can sometimes lead to mismatch in `scrapers-ca`, but the Heroku app will display an error if so).

`province-nl-electoral_districts.csv` lost an electoral district, which it shouldn't have. Looking at the source, http://www.assembly.nl.ca/members/cms/membersalpha.htm, it's because that seat is vacant. I recommend simply reverting the change to `province-nl-electoral_districts.csv`.

Also, don't forget to run `./scripts/compile.py ca`. This step should be added to the Maintenance section of the readme. https://github.com/opencivicdata/ocd-division-ids/blob/master/scripts/country-ca/README.md Also, we should add some text there about manually checking whether the diffs of the CSVs make sense. For example:
* If a row is added, check that a new boundary has in fact been added to that jurisdiction, e.g. by visiting the website of the chief electoral officer.
* If a row is deleted, check that a boundary has in fact been removed from that jurisdiction, or that the row exactly duplicates another row.
* If a row is changed (usually the boundary name), check that the new name makes sense.

jpmckinney:
Also, please add the link http://open.canada.ca/data/en/dataset/5f1c2e06-405a-4357-9566-872e69ee2ade as a comment in the code above the shapefile link (this is done elsewhere in the code).

cmyr:
@jpmckinney Alright, I think I've gotten everything you mentioned.

jpmckinney:
Thanks!

PR #152, opened 2018-02-22
Add UK electoral divisions and places

showerst:
Sourced from UK ordnance survey Electoral boundaries and Named Places datasets. 

See [README](https://github.com/showerst/ocd-division-ids/tree/uk/scripts/country-uk) for additional documentation on new types, and source and explainer links.


chris48s:
Hello. First off, apologies for the drive-by review. I appreciate that I have no previous connection to the project and am just diving in with _opinions_ about _stuff_.


**Why do I care about this?**

I work on a number of UK-based projects where we publish open data on election [candidates](https://candidates.democracyclub.org.uk/help/api) and [results](https://candidates.democracyclub.org.uk/help/results). We're interested in this pull request as potential consumers of OCD division IDs because we are looking at using Google's Election Common Standard Data Specification as a format for publishing results data. If IDs are added for the UK, it is likely we would be the first users of these IDs in the relevant jurisdiction.

Looking over this PR, I can see that you have put a lot of work into this, done research into the problem domain and accounted for a number of non-obvious edge cases. Importing names from BoundaryLine is an obvious place to start, but there are a number of areas where I have some questions to ask or where I may be able to offer some local knowledge which might be helpful (or possibly just convince you that the processes for publishing electoral division data in the UK are completely ridiculous).


* Local electoral divisions are routinely reviewed and frequently changed but Ordnance Survey data is generally published retrospectively. A new or revised boundary may have been introduced via legislation, but it will only appear in BoundaryLine after an election has taken place.

  To give one worked example of this (I'll try not to drown you with excessive detail here):

    * Hertsmere local authority has recently had a boundary review, changing its wards (these will become 'District Wards' - `DIW` in Boundary Line). The new names and boundaries for the divisions are known and the legislation which enacts the change has been passed: http://www.legislation.gov.uk/uksi/2018/535/contents/made
    * The first local election using these new boundaries will occur on 2nd May 2019
    * Although the new boundaries and names are known and you could (in principle) create future-dated OCD identifiers for them today, they have not yet been published in any OS BoundaryLine release.
    * The first OS BoundaryLine release to contain them will be the May 2019 release which will usually be published after the first elections using those boundaries have already been contested.
    * This is common practice and will affect some subset of local division identifiers in every election. For example this will be the case for 677 electoral divisions in the local elections scheduled for May 2019.

  This means that creating identifiers for local electoral divisions using only BoundaryLine as an input source will create a set of identifiers which are suitable for identifying divisions in many cases, but there will some subset of local elections where it would not possible to use OCD division ids to publish data about candidates in upcoming elections or to publish results data in a timely fashion.

  In general, this is a problematic area for organisations working on elections data in the UK but we are working on solutions. For example, we provide [a python package for parsing division names from Electoral Change Orders](https://github.com/DemocracyClub/eco-parser/) into machine readable data.

* Additionally, OS BoundaryLine doesn't cover the whole of the UK. Ordnance Survey publishes data for Great Britain (England, Wales and Scotland) but not Northern Ireland. This means the set of divisions for the UK Parliament proposed in this PR is incomplete. The Westminster parliament has 650 representatives each representing a constituency but this file only contains 632 constituency identifiers because the Northern Ireland constituencies are missing. Also the local/regional structures for Northern Ireland are missing, but that is another issue.

* There are also some other areas in England and Wales which are definitely "political geographies" with elected representatives (e.g: combined authority mayors) which aren't published in OS BoundaryLine.

* The decision to create division identifiers for ceremonial counties seems slightly odd. Reading the definitions in http://docs.opencivicdata.org/en/latest/proposals/0002.html I don't think ceremonial counties meet it. The County Electoral Divisions (`CED`s are divisions of Administrative Counties as opposed to Ceremonial Counties).

* GSS codes are very useful identifiers. They are rigorously maintained by the Office for National Statistics, widely used and essential for data interoperability with all official statistics published in this country. It should be noted they identify a _boundary_ as opposed to a _division_ and will change even with minor 'technical' boundary changes. Your [specification](http://docs.opencivicdata.org/en/latest/proposals/0002.html) states that an identifier maps to a division as opposed to a boundary. As such, to present a division id as having a one-to-one mapping with a GSS code is not quite nuanced enough. To give one example, your proposed division identifier `ocd-division/country:uk/uta:sir_gaerfyrddin_-_carmarthenshire/ute:llangennech` could describe [W05000492](https://mapit.mysociety.org/area/12665.html) or [W05001003](https://mapit.mysociety.org/area/151560.html) at different points in time. How could we represent this relationship more accurately within the conventions of the project?

* Additionally there are some electoral divisions which do have GSS codes which are published but aren't published in BoundaryLine (yet). For example, the Office for National Statistics do [publish GSS codes for County Electoral Divisions](http://geoportal.statistics.gov.uk/datasets/county-electoral-division-may-2017-names-and-codes-for-england) but they're not in BoundaryLine.

* I'd like to understand more about the process for managing change over time in OCD identifiers. The files in this PR are generated by running these scripts against some BoundaryLine release at a point in time, but if you run them againt another release a bunch of different files are going to pop out. In order to generate a set of identifiers which is useful for more than one instant in time, some thought needs to go into how to reconcile and manage change. In general, being able to specify the range of time for which an identifier is valid (with reference to relevant Electoral Change Order legislation) is important for dfining which set of identifiers is relevant for a given election.

More generally, it is useful that there is interest in adding UK identifiers to the project, we're interested in using them but have some concerns that the approach in this PR potentially formalises a set of IDs which may not meet the needs of consumers in the relevant jurisdiction. We'd like to explore ways we might be able to help. Simultaneously, there are some issues I've raised here relating to local identifiers where we are still finding our way, or where the data landscape is changing. Identifiers relating to national-level elections seem to be the area where there is the most appetite for data to be published using OCD identifiers and hence that subset would seem a good place to start. What would be a good next step?


/cc @symroe

identifiers/country-uk/district_borough_unitary_ward.csv
```diff
@@ -0,0 +1,7306 @@
...
+ocd-division/country:uk/cty:devon/diw:buckland_monachorum_ward,Buckland Monachorum Ward,E05010552,
+ocd-division/country:uk/cty:devon/diw:milton_ford_ward,Milton Ford Ward,E05010560,
+ocd-division/country:uk/cty:devon/diw:mary_tavy_ward,Mary Tavy Ward,E05010559,
+ocd-division/country:uk/cty:devon/diw:chagford_ward,Chagford Ward,E05010554,
+ocd-division/country:uk/cty:devon/diw:okehampton_south_ward,Okehampton South Ward,E05010562,
+ocd-division/country:uk/cty:devon/diw:south_tawton_ward,South Tawton Ward,E05010563,
+ocd-division/country:uk/cty:devon/diw:okehampton_north_ward,Okehampton North Ward,E05010561,
+ocd-division/country:uk/cty:devon/diw:tavistock_north_ward,Tavistock North Ward,E05010565,
+ocd-division/country:uk/cty:devon/diw:tavistock_south_west_ward,Tavistock South West Ward,E05010567,
+ocd-division/country:uk/cty:devon/diw:tavistock_south_east_war
```
chris48s: There's an edge case which is not handled here in the parsing code. Note that `Bisley Ward (DET)` and `Bisley Ward` (2 rows up) have the same GSS code (`E05007376`). That is because they're both parts of the same division. (on a map: https://mapit.mysociety.org/area/E05007376.html ). Quote from the BoundaryLine guidance:

> The description ‘Detached, (Det)’, only applies to portions of local government or parliamentary constituency areas which are separated from the main area, being completely surrounded by other local government or parliamentary constituency areas and not connected by direct access on the ground.
> 
> The description will not be applied to islands or parts of islands in the sea.
> 
> Detached parts have the same AI attribute (and GSS code where applicable) as the main area..

so `Bisley Ward` should only have one OCD division identifier. There's a number of other occurrences of this same condition.

scripts/country-uk/ordnance.py
```diff
@@ -0,0 +1,288 @@
...
+                          r'_district$',
+                          r' Authority$',
+                          r'_authority$',
+                          ]
+    for replacement in regex_replacements:
+        type_id = re.sub(replacement, '', type_id, flags=re.IGNORECASE)
+
+    type_id = type_id.lower()
+    type_id = re.sub('\.? ', '_', type_id)
+    type_id = re.sub('[^\w0-9~_.-]', '~', type_id, re.UNICODE)
```
chris48s: I'd like to understand more about the slugging algorithm used here. To what extent is this mandated by the spec/project convention/individual preference?

showerst:
@chris48s thanks so much for the read-over -- I'll take a look at all your feedback this weekend. Definitely open to improvements here.

scripts/country-uk/ordnance.py
showerst: [Pulled from the spec](http://docs.opencivicdata.org/en/latest/proposals/0002.html#identifier-scheme). I believe I took the regex from [elsewhere in the project](https://github.com/opencivicdata/ocd-division-ids/blob/master/scripts/country-us/census_places.py#L167) but it's been a few months so i'm not sure.

The replacements were an attempt not add redundant data (to avoid things like ocd-division/country:uk/county:sample/sample_county) and to smooth out some stylistic choices that seemed to be Ordnance specific.

If you catch any cases where i'm grossly mishandling special characters or other abbrevs, or this ID scheme is doing something that makes no sense to a local user, please let me know.

identifiers/country-uk/district_borough_unitary_ward.csv
showerst: Looks like we need a special case for that.

showerst:
PS @chris48s my email is in my profile, feel free to shoot me a message if you have questions / ideas that might not fit this PR.

chris48s:
Thanks very much for this info. I'll have a proper read through it all and aim to respond later in the week.

chris48s:
> the USA, home of the non-contiguous multi-polygon

..and the land of the brave. :us:  Pretty sure those are the words, yeah? :D

Thanks for all the additional detail here. Now that I've had a bit more time I've read and fully digested all of this. There's lots of interesting stuff to get into the weeds on here, and I have started drafting replies to some of it, but before getting into a lot of detail on some of these topics (which may not be the most useful thing to spend a lot of time at this stage) there's one thing I'd like to understand more about:

> congressional districts occasionally get re-numbered and re-coded, so "FL 9th House District" might move by hundreds of miles and be represented by a totally different person. This is OK, as the goal of the project is to create a standardized unit for "FL 9th House District", while "the house district that covers Orlando as of 2018-01-01" is slightly orthogonal in scope

> As stated above with my Florida example, the idea is to have a stable ID for a given political unit over time, even if that political unit "moves" it's location/boundaries. FL 9th is always FL 9th, no matter where in the state it is

> This makes sense in a US context, because when you're discussing "the mayor of Phoenix" or "The person who represented the FL 9th House District in 1996" you care more about the role than the specific geography. I'm not sure how well this translates to the UK context.

This stability is a useful property at some level. Similarly to your Phoenix example, "The Mayor of Bristol" is a meaningful title, regardless of whether the city of Bristol experiences minor boundary changes over time.

Unfortunately I suspect this concept probably exports poorly to local government divisions where the geographic area represented and number of representatives it elects is more important than the name attached to it. In fact, local wards can change their name fairly trivially. It is a decision taken at local level and renaming an area doesn't even change its GSS code. Conversely, for an area to re-use a previously used name but substantially change the area it covers is a change that requires legislation to enact and forces an early election in the area in order to enact this change, even if one is not scheduled.

It seems like in your example, the name "9th House District" is conveying semantic meaning that is not encoded in a name like "Marple South and High Lane ward" and it seems like trying to translate that concept will create an identifier which changes when something really trivial changes, but can generate the same 'stable' identifier for what we would consider to be fundamentally different divisions at different points in time (but which share a common name within a scope). This seems like the balance of stability/volatility is the wrong way round for this situation.

I wonder if attempting to assign an OCD identifier is actually a useful thing to do in that context?

I don't have much concept of how this maps on to other countries. I wonder if there is another example of a country already covered by this project which is more similar? I also notice having a very quick browse through the repo that for a lot of countries, only relatively high-level identifiers exist. Maybe this is for similar reasons?

___


There are a couple of things I'll get into some more detail on now as this is probably (hopefully) useful information regardless of the above. I'm going to use the confusingly similar abbreviations OS (Ordnance Survey) and ONS (Office for National Statistics) a lot in this next section...

## Data Source

> To be honest I don't love using the Ordnance source since it requires a one-time email-gated download and parsing shapefiles, but I researched for a while and didn't find anything else that was close to comprehensive while also being official and free. Happy to be directed to better sources.

If we care about _boundaries_, Ordnance Survey data is necessary evil. Very high quality necessary evil. It can be a bit of a pain to deal with for various reasons, but fundamentally it is comprehensive (within scope), extremely well maintained, etc. It is worth noting that mySociety helpfully mirror the data at http://parlvid.mysociety.org/os/ . If you grab it from there, you don't have to go through the whole "add to cart, checkout, pay £0" dance to get the data.

However, if all you care about is _names and codes_ we might want to look at using data from the Office for National Statistics instead. The ONS offer (amongst other things) much simpler CSV downloads. e.g:

Westminster Constituencies:
- http://geoportal.statistics.gov.uk/datasets/westminster-parliamentary-constituencies-december-2017-full-extent-boundaries-in-the-uk-wgs84
- https://opendata.arcgis.com/datasets/5c582cef61d04618928639dd17e4f896_1.csv

Welsh Assembly Constituencies:
- http://geoportal.statistics.gov.uk/datasets/national-assembly-for-wales-constituencies-december-2017-wa-bfe
- https://opendata.arcgis.com/datasets/7a4dc40d6c9e4abf8d2fa3b369395d93_1.csv

Welsh Assembly Regions:
- http://geoportal.statistics.gov.uk/datasets/national-assembly-for-wales-electoral-regions-december-2017-wa-bfe
- https://opendata.arcgis.com/datasets/d10026914aa64c4cadcfafad73f81bf7_1.csv

etc etc.


All this stuff is available from http://geoportal.statistics.gov.uk/ although it can be a bit difficult to navigate the absolute kitchen sink of open data available here (I'm happy to help with this). Naturally there are some tradeoffs to think about here.

Pros:

* ONS do _sort of_ cover Northern Ireland. (This is complicated, but they at least publish NI data to a greater extent than OS do. If you grab a dataset that you would expect to cover the entire UK like 'Westminster Constituencies', it actually covers the entire country rather than just part of it).
* ONS data abstracts a lot of the edge cases you have to deal with in parsing BoundaryLine. They deal with a lot of the things that are going to trip you up and package it in a more user-friendly way.
* The CSVs from ONS are much closer to what you're trying to generate as an end result than the OS data.
* ONS publish names/codes for some of the area types I mentioned which have elected officials but aren't in OS BoundaryLine (e.g: Combined Authorities for Combined Authority Mayors and Police Areas for Police and Crime Commissioners)

Cons:

* The publication schedule for ONS data is generally further behind than BoundaryLine. Often quite a lot further behind. This makes all the problems noted above about changes in legislation not being reflected in data in a timely fashion even worse than if you use OS as a base. Simultaneously, either way you'll need some kind of process to deal with making changes before they appear in published data.
* ONS datasets are non-hierarchical. If you limit the scope to only relatively high-level identifiers this isn't a problem, but if you want to get into identifiers for local government divisions (wards, county electoral divisions, etc) you need some way to build up the hierarchy. so for example, if I grab this https://opendata.arcgis.com/datasets/fba403b550d3456b813714bfbe7d0f0c_1.csv it tells me there are 24 wards in the country called 'Abbey' but not what local authority each one is in, so you would still need some additional processing to create them as subdivisions of the right local councils. I guess this is kind of like knowing there are 50 "9th districts" and having an identifier for each one but not knowing which state each one is in? (except that "Abbey" is just a very common name rather than something that conveys semantic meaning).
* ONS abstract a some detail about _exactly what type_ of division this is. For example, their "wards" data doesn't attempt to address the issue of whether this is a `UTW` (unitary ward) or a `DIW` (district ward) or a `MTW` (metropolitan borough ward) etc - they're all just wards. I note that you've chosen to encode this info in the IDs in this PR, but it may be that is more a symptom of how Ordnance Survey organise it in BoundaryLine rather than because this is strictly necessary.

## GSS codes

> Is a GSS code mappable to one division (or boundary wholly contained by a division) at any single point in time (aka it's a one to one relationship that just changes often, vs a potential one to many)?

Yes - a division id can map to a GSS code and a range of dates. It shouldn't ever have 2 different GSS codes _at the same time_. In most cases that range of dates is fairly straightforward to infer. In some edge cases working out the exact date ranges may be tricky, but lets abstract that detail for now..


___


Finally, no rush to respond to any of this. I'm not going to be working next week, so am unlikely to respond to any reply this side of the week starting 13th August although it is possible my colleague @symroe might have some input on this next week. Hope this helps.

jpmckinney:
I can answer some of the philosophical questions about OCIDs.

Bizarre logics often operate around political divisions, because the goal of districting isn't to produce a clear, orderly geographic system, but to relate a decision-maker/body with a jurisdiction. Similarly, although some people try to use postal codes as proxies for location identifiers, the goal of postal codes is primarily to support the delivery of mail, not the identification of locations, and those people quickly run into trouble.

So, in trying to have a clear, orderly identification system for political divisions, we inevitably face challenges, because they were not designed to facilitate such identification.

OCDIDs identify political divisions, whose names and geographies may change entirely. Deciding whether something changes the identity of a division is up to the maintainers of each country.

For example, in Canada (one of the countries with many identifiers, which I maintain), a redistricting at the federal level will abolish districts, create districts, rename districts, and redraw districts. Why do they sometimes abolish instead of renaming/redrawing? It's a mystery. A fastidious maintainer could respect this bizarre logic and have identifiers 'survive' renaming/redrawing but not abolishment. But Canada, sadly, offers no way to do so (they change names and re-assign identifiers) except by reading the full text of a Representation Order. So, in practice, I mint new identifiers with each redistricting, adding a suffix to the source identifier for the year of the Representation Order, and using `validFrom` and/or `validThrough`.

For other jurisdictions, I attempt to mint an identifier that omits the name, which can change more frequently; for example, some localities both number and name wards.

For minor jurisdictions, that have no data management, I currently only keep present, not historical, identifiers.

Some of these decisions may sound hair-raising to some, and reasonable to others. It depends on the use cases that the community of users are trying to support.

In the US, although Florida's 9th congressional district may be an entirely different geography in 2008 and in 2018, it serves the community to treat it as the same division across time.

There's more information on [identifier governance here](http://docs.opencivicdata.org/en/latest/proposals/0008.html).

symroe:
This is a great thread and I don't think I can add a lot, other than:

I think there is an inverse correlation between the "level" of division and public interest (and therefore likelihood of OCDID use). That is, parliamentary elections have stable IDs that make headlines when they change, [new data is published well in advance](https://www.bce2018.org.uk/publications), and are likely to be of interest to a lot of people. They also change infrequently. 

Can I suggest we limit the scope of this PR by picking a few types of divisions that meet this profile and then adding the more complex cases after? This is simply so that the UK has _some_ coverage, especially in the context that we might have a snap general election in the next few months.

A reasonable subset might be: Police, Parl, Scottish, Northern Ireland and Welsh parl, CA-Mayors – see note above about the 'composite areas' in the case of PCC and Mayors though.

Of course we could work on all the complex cases at once in this PR, but tracking everything might get tricky.

In terms of tooling around changes, we have some bots that scrape UK legislation and the boundary change website and notify us of changes. I expect we could find a way to share that in future, or simply work on a PR at the same time as we work on adding the data to our own tools. If you're interested in this (and other UK election data), you might want to [join our Slack](http://slack.democracyclub.org.uk/) – although this thread is obviously the best place to talk about this exact work.

chris48s:
Thanks for all the additional info on this. I think probably the most important point for me is

> It depends on the use cases that the community of users are trying to support

I'm totally on-board with that.

As noted in Sym's response, the divisions where there is most interest in consuming data attached to a OCD identifier are those which are relevant to national-level politics (which also correspond to those which change the least frequently and in a more high-profile way). It is currently somewhat difficult to evaluate right decisions/tradeoffs to make about divisions relevant to local government when the use-cases are abstract or not understood yet.

I'd be happy to take this forward by working on a PR to add identifiers (and associated scripts) for the subset suggested by @symroe based on the ONS data. Does that sound like a reasonable next step?

chris48s:
Cool. I'll aim to slot in some work on this later this week or next week :)

chris48s:
#156 now open and ready for review

showerst:
I'm closing this in favor of https://github.com/opencivicdata/ocd-division-ids/pull/156 which is better thought out.

PR #153, opened 2018-02-23
IE: add base, upper and lower parliament divisions

showerst:
Add Ireland, and its legislative subdivisions.

### DIFF
--- a/identifiers/country-ie.csv
+++ b/identifiers/country-ie.csv
@@ -1,0 +1,44 @@
+id,name
+ocd-division/country:ie,Republic of Ireland
+ocd-division/country:ie/constituency:athlone~longford,Athlone–Longford
+ocd-division/country:ie/constituency:carlow~kildare,Carlow–Kildare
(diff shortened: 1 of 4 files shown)


showerst:
Looks like the API i was using was missing some, or basing on older data. Closing this for now until I get that cleared up.

showerst:
I fixed the issue -- the Oireachtas API wasn't returning the latest data by default, so it just needed additional flags.

Ireland has some unusual practices in its [upper house](https://en.wikipedia.org/wiki/Seanad_%C3%89ireann), some members are nominated by graduates of major universities, and some by practitioners in certain fields. A member can be said to be representing "National University of Ireland" panel though, and that's how its they're referred to on [the website](http://www.oireachtas.ie/members-hist/default.asp?housetype=1&HouseNum=25&disp=mem). 

I'm not quite sure how to model these, they're not really divisions _or_ jurisdictions per-se, but they're political divisions represented by legislators, so they seem to fit the project. I'm open to revisions there.

PR #156, opened 2018-08-19
Add UK Electoral Divisions

chris48s:
This follows on from the conversation started in #152

**Done in this PR:**

I've started off by adding scripts to create ids from ONS data for:

* UK Parliament Consitituencies
* Scottish Parliament Consitituencies
* Scottish Parliament Regions
* National Assembly for Wales Consitituencies
* National Assembly for Wales Regions
* Combined Authorities
* Police Force Areas

I look forward to review/feedback on this.

**Possible Future Work:**

1. Ongoing maintenance

Inevitably at some point there will be changes to these divisions which we need to account for. The next substantial legislative change to impact these divisions is likely to be this review of UK Parliamentary Consitituencies:

* https://boundarycommissionforengland.independent.gov.uk/2018-review/
* https://en.wikipedia.org/wiki/Sixth_Periodic_Review_of_Westminster_constituencies

I'm happy to keep on top of this moving forwards.

2. GSS Code changes

At the moment everything has a _current_ GSS code. These boundaries are relatively stable and change quite infrequently, but many of the divisions in this dataset have had boundary changes at some point and hence it is useful to be able to map a division to >1 GSS codes based on a date range.

I have not attempted to deal with that here but I may look at using the [Code History Database](https://www.ons.gov.uk/methodology/geography/geographicalproducts/namescodesandlookups/codehistorydatabasechd) to do this as future work.

3. Add more division ids

As discussed in #152 there is a large body of division identifiers (mainly relevant to local elections) which have some additional complexity. Lets review the best way to manage them in light of future interest/use-cases.

### DIFF
--- a/scripts/country-uk/data.py
+++ b/scripts/country-uk/data.py
@@ -1,0 +1,89 @@
+# The UK Parliament (House of Commons) has members at consitituency level
+# in England, Wales, Scotland and Northern Ireland
+pcon = {
+    # data from: http://geoportal.statistics.gov.uk/datasets/westminster-parliamentary-constituencies-december-2017-full-extent-boundaries-in-the-uk-wgs84
+    'url': 'https://opendata.arcgis.com/datasets/5c582cef61d04618928639dd17e4f896_1.csv',
+    'gss_column': 'pcon17cd',
+    'name_column': 'pcon17nm',
+    'prefix': 'pcon',
+    'filename': 'uk_parliament_consitituencies.csv'
+}
+
+
+# The Scottish Parliament has members at both consitituency and regional levels
+spc = {
+    # data from: http://geoportal.statistics.gov.uk/datasets/scottish-parliamentary-constituencies-may-2016-full-extent-boundaries-in-scotland
+    'url': 'https://opendata.arcgis.com/datasets/00436d85fa664f0fb7dce4a1aff83f27_1.csv',
+    'gss_column': 'spc16cd',
+    'name_column': 'spc16nm',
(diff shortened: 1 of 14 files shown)


scripts/country-uk/data.py
```diff
@@ -0,0 +1,89 @@
+# The UK Parliament (House of Commons) has members at consitituency level
+# in England, Wales, Scotland and Northern Ireland
+pcon = {
+    # data from: http://geoportal.statistics.gov.uk/datasets/westminster-parliamentary-constituencies-december-2017-full-extent-boundaries-in-the-uk-wgs84
+    'url': 'https://opendata.arcgis.com/datasets/5c582cef61d04618928639dd17e4f896_1.csv',
```
chris48s: Would it be better to commit the source CSV files here and document where they came from instead of linking them directly so that this process is more "reproducible". I can see reason for going either way on this. Happy to take guidance.

chris48s:
Hi. Any update on this? It would be really useful to get a review from a maintainer on this.

</polite bump>

jpmckinney:
@showerst @symroe May I request your review?

showerst:
Sure, i'll check this over this weekend.

showerst:
@jpmckinney this all built fine, and the output looks solid to me.

jpmckinney:
Thanks!

symroe:
🎉 Excellent – I'd just started reviewing (I work with Chris so knew some of what he'd done already). All looks great, and thanks for merging!

PR #159, opened 2019-03-27
+Fremont County WY council districts

rsimoes:


jdmgoogle:
The changes to the WY file look good, but do you know why GitHub thinks there are changes to the other 8,010 lines in the country-us.csv file?

large_diff

jpmckinney:
It's possible that newline characters were changed.

rsimoes:
For some reason the country-us.csv file had a mix of `\n` and `\r\n` breaks. I'm not sure how that would have been possible without prior manual editing.

jdmgoogle:
What do they have now?

jpmckinney:
@jdmgoogle If compile.py was run, then I assume this PR has the intended line terminators. (This can be verified by making a local copy of the branch and re-running it.)

rsimoes:
CRLF, which is in line with RFC 4180. I think that's the default setting for Python's core `csv` library.

jdmgoogle:
SGTM. Approved.

PR #161, opened 2019-06-19
Update ocdids for Canada and Japan

riturajj:
1. Update OCDIDs for Canada.
2. Add top level OCDID for Japan


jpmckinney:
Hmm, can you explain the reasons for the changes to Canada?

riturajj:
The identifiers/country-ca.csv file included federal electoral district ids that came from 2 csv files - ca_federal_electoral_districts-2013.csv and ca_federal_electoral_districts.csv. The ones in *-2013.csv were the updated ones (338) so replaced them with ca_federal_electoral_districts.csv (308). I think instead of having ids from both csv files, we can have from one.

jpmckinney:
You can filter for the new IDs using the `validFrom` column. Historical use cases need access to the old IDs.

jpmckinney:
You can re-submit a separate PR for Japan, but change the country code to `jp` (according to ISO-3166-1).

PR #162, opened 2019-06-20
Add top level OCDID for Japan

riturajj:
Add top level OCDID for Japan


jpmckinney:
I approve. Waiting for second approval based on http://docs.opencivicdata.org/en/latest/proposals/0008.html#general-contributions

Issue #170, opened 2019-09-04
Update/establish best practices for OCD-ID types across countries


jdmgoogle:
This issue spun out of the discussion on PR #168.

Background links:
[Creating new OCD-IDs](https://opencivicdata.readthedocs.io/en/latest/ocdids.html)
[Division identifiers](http://docs.opencivicdata.org/en/latest/proposals/0002.html)

The current OCD-ID documentation establishes one canonical identifier type -- `country` -- but otherwise gives latitude to within-country maintainers to define and enforce types appropriate for their jurisdiction. E.g., the first-level administrative type in the U.S. is a `state`, in Portugal it is a `district`, and in Germany it is a `land`.

The current situation gives a lot of flexibility and discretion to these within-country maintainer, but can place a significant burden on consumers of the identifiers to figure out what common types are across countries; e.g., is a `district` a first-level administrative division, or a sub-city legislative division?

Previous ad-hoc attempts to address this problem (e.g., PR #148) created identifiers which used the in-country types as aliases of identifiers that were more American in origin. E.g.,

Canonical: `ocd-division/country:de/state:bb`
Alias: `ocd-division/country:de/land:bb`

This was an attempt to balance the needs of publishers (using in-country types and terminology) and consumers (using types they had already seen). The discussion in PR #168 came to the conclusion that this was not desirable, or at least should go through a more thorough review before being implemented at scale. The options discussed were:

1. Codify the ad-hoc approach.
2. Reverse the current ad-hoc approach, in that the alias should be from the local term to the across-country term.
3. Follow the practices of projects like GeoNames, etc., which use country-agnostic terms like `adm1`, `adm2`, etc, and not use the US-centric terms `state` and `cd` for a global specification.
4. Add a new file to the spec, e.g. `country-de-types.csv`, with the columns local-type and standard-type, with rows like `land,adm1` and `wahlkreis,constituency`.

Discuss. :)

jpmckinney:
Using the local terms will always make sense for local users.

A challenge only arises when a user (1) wants to use OCDIDs from multiple jurisdictions and (2) needs to know whether a type ID in one jurisdiction corresponds to a type ID in another jurisdiction. (@jdmgoogle Can you provide a specific use case?)

Now, the problem of deciding whether two divisions in two jurisdictions are of the same type is not a solved problem – by anyone, anywhere.

[...]

PR #172, opened 2019-09-07
Add top level OCDID for Netherlands / existing Parliament (Tweede Kamer)

sguenther85:


jdmgoogle:
LGTM

jdmgoogle:
Although please update the PR description since this only includes the top-level OCD-ID, and not the parliamentary districts (which are blocked on the resolution of #170).

jpmckinney:
In some cases we use the English name (Taiwan), in others we use a local name (Polska, Nederland). Do we have a preference, or is it whatever the contributor proposes?

sguenther85:
I would prefer the local names.
But in cases of Chinese, Korean or Japanese I would suggest the alphabetical variant of the language.
For Example:

Englisch: Japan
Japanese: 日本
Japanese: Nippon

Much better would be a second columnin the .csv File to set the local name and a possible english translation for everybody.

sguenther85:
@jdmgoogle but there are no political districts/region or what else for last netherlands election. 
And i can't just add the first level districts without knowing if the geographical districts are also the political districts for elections in the netherlands?

https://en.wikipedia.org/wiki/2017_Dutch_general_election
The House of Representatives (Dutch: Tweede Kamer) is composed of 150 seats elected by proportional representation in a **single nationwide constituency**, with a legal threshold of 1 full seat (0.67%), and residuals assigned by the D'Hondt method.[6][7] The Senate is indirectly elected by the States-Provincial.

sguenther85:
@jdmgoogle can we approve and merge this PR or is there anything todo? Like i wrote last time: They have only on single nationwide constituency!

PR #173, opened 2019-10-01
ocd for 'November 2019 Spanish general election' and existing parliament

sguenther85:


identifiers/country-es.csv
```diff
@@ -0,0 +1,67 @@
+id,name
+ocd-division/country:es,España
+ocd-division/country:es/autonomous_region:ce,Ceuta
+ocd-division/country:es/autonomous_region:ml,Melillaocd-division/country:es/state:an,Andalucíaocd-division/country:es/state:an/cd:almería,Almeríaocd-division/country:es/state:an/cd:cádiz,Cádiz
```
jdmgoogle: Looks like there are some multi-line issues in the merging.

identifiers/country-es/ciudad_autonoma.csv
```diff
@@ -0,0 +1,2 @@
+ocd-division/country:es/autonomous_region:ce,Ceuta
+ocd-division/country:es/autonomous_region:ml,Melilla
```
jdmgoogle: I'd lean towards "region" here instead of "autonomous_region" but I won't block on that.

identifiers/country-es/comunidad_autonoma.csv
```diff
@@ -0,0 +1,19 @@
+id,name
```
jdmgoogle: I think based on where the discussion is going in #170 we should use something like "community:" in place of "state:" here, but I'd like to get @jpmckinney's thoughts.

identifiers/country-es/provincia.csv
```diff
@@ -0,0 +1,50 @@
+ocd-division/country:es/state:ga/cd:a_coruña,A Coruña
```
jdmgoogle: Similarly because of #170 something like "province:" instead of "cd:" makes more sense. @jpmckinney

sguenther85: don't have a problem with province instead of cd. would be okay with that.
But should i use then 
- ocd-division/country:es/state:ga/province:a_coruña,A Coruña
OR with hasc code (because it isnt a cd)?
- ocd-division/country:es/state:ga/province:ac,A Coruña

identifiers/country-es/ciudad_autonoma.csv
sguenther85: i used "autonomous_region" because of the portugal PR.
Where we have a difference between i.e Lisboa and Acores

ocd-division/country:pt/region:li - Lisboa
ocd-division/country:pt/autonomous_region:ac - Acores
Lisboa is a region in Portugal, but Acores is a autonomous region (no normal region).

So I'd suggest to use "autonomous_region" in this special cases.

identifiers/country-es.csv
sguenther85: Should be fixed now

identifiers/country-es/comunidad_autonoma.csv
sguenther85: aliases would also be possible?

ocd-division/country:es/community:ga/
"same as"
ocd-division/country:es/state:ga/

And yes, spain has in this case no states, but "autonomous communities".
What do you think?

sguenther85:
@jdmgoogle @jpmckinney 
it is possible to continue here?

@jpmckinney any question or opinions about @jdmgoogle and my questions above?
@jdmgoogle any opinions on my las comment to your review-changes?

identifiers/country-es/ciudad_autonoma.csv
jdmgoogle: SGTM.

identifiers/country-es/comunidad_autonoma.csv
jdmgoogle: SGTM (minus the trailing `/`s, of course).

identifiers/country-es/provincia.csv
jdmgoogle: If there's an official set of abbreviations (either ISO or HSAC) for provinces, then use that. Otherwise the full names are good.

Also, could we sort these by OCD-ID instead of name? It's easier to look it over and see how many provinces are in each state.

identifiers/country-es/comunidad_autonoma.csv
sguenther85: Yes, without / at the end. My fault.

identifiers/country-es/provincia.csv
sguenther85: Yes, there is. http://www.statoids.com/yes.html
ocd-division/country:es/state:ga/province:a_coruña,A Coruña
HASC:  ES.GA.**AC**

And yes, i can sort by ocd-id. should be no problem.

identifiers/country-es/comunidad_autonoma.csv
sguenther85: I used now autonomous_community instead of state. And changes the code from hasc to iso fpr the autonomous_communitie's.
@jpmckinney @jdmgoogle 
Anything else here to do?

identifiers/country-es/provincia.csv
sguenther85: cd changes to province with iso code
@jdmgoogle @jpmckinney

identifiers/country-es/ciudad_autonoma.csv
sguenther85: I changes both now to "autonomous_city". 
Sorry for confusion. I used portugal as an example for spain, because of the similar problem.
Ceute and Melilla have in spain an other "political status". So my suggestion would be to have a difference to the other communities, but if you both say "we should use the same ocd type", I can change that again.
@jpmckinney @jdmgoogle

jpmckinney: I don't know the details for Portugal, so I defer to you, @sguenther85.

PR #174, opened 2019-10-03
Create 2019 Indian Vidhan Sabha OCDIDs.

rahul-nath:
In this PR I've included a script that I created to generate OCD IDs specifically for the Indian Vidhan Sabha elections of Maharashtra and Haryana. The OCD IDs generated are for constituencies of these states, which include their districts. ~There were some decisions made regarding district~ ~names due to discrepancies between the districts in wikipedia pages for [Maharashtra districts]~(https://en.wikipedia.org/wiki/List_of_districts_of_Maharashtra#Districts) and [constituencies]~(https://en.wikipedia.org/wiki/List_of_constituencies_of_the_Maharashtra_Legislative_Assembly)~ ~and [Haryana districts](https://en.wikipedia.org/wiki/List_of_districts_of_Haryana#Districts) and~  
 ~[constituencies]~
~(https://en.wikipedia.org/wiki/List_of_constituencies_of_the_Haryana_Legislative_Assembly). The~ ~district pages were deferred to over the analogous columns in the constituency pages after some~ ~research. They are as follows:~

~- `Yamunanagar` is used over `Yamuna Nagar`~
~- `Gondia` is used over `Gondiya`~
~- `Gurugram` is used over `Gurgaon`~
~- `Nuh` is used over ``Mewat~

~where applicable.~

UPDATE: For source of truth, it was determined that `https://affidavit.eci.gov.in` is the source of truth regarding Consituency and District names.

Additionally, no changes were made to the aliases file located in `identifiers/countries-in` as it's unclear if that was necessary.


rahul-nath:
@jamesturk @jpmckinney @jdmgoogle this is necessary for the imminent Vidhan Sabha elections. I am unable to assign reviewers so please assign yourself.

rahul-nath:
Just want to bump this to make sure it's been seen.

jpmckinney:
Format looks good to me – I haven't checked against source of truth.

scripts/create_ocd_ids.py
```diff
@@ -0,0 +1,144 @@
...
+          "id": ocd_id.format(country, row["state_abbr"], row["district_abbr"], row["constituency"]),
+          "name": rest.format(row["state"], row["district"], election, full_const)
+      }
+      parent = ocd_id_row["id"].rsplit("/", 1)
+      parent_set.add("/".join(parent[:-1]) + "," + row["district"])
+      writer.writerow(ocd_id_row)
+  for parent in sorted(parent_set, key=lambda x: x.split(",")[-1]):
+    print(parent)
+
+for state_abbr, state  in contests.items():
```
jdmgoogle: Please sort the output so it's easier to read.

scripts/create_ocd_ids.py
```diff
@@ -0,0 +1,144 @@
...
+    c_row["state_abbr"] = state_abbr
+    c_row["district"] = cons_district
+    new_table.append(c_row)
+  return new_table
+
+def write_to_file(table):
+  # format hardcoded OCD ID
+  global new_file
+  ocd_id = "ocd-division/country:{}/state:{}/district:{}/cd:{}"
+  rest = "state {} district {} {} constituency {}"
```
jdmgoogle: The naming here is a bit awkward. Maybe

${constituency} constituency, ${district} district, ${state}

E.g.,

Khanapur constituency, Sangli district, Maharashtra

scripts/create_ocd_ids.py
```diff
@@ -0,0 +1,144 @@
...
+    "Kalanwali": "Kalawali",
+    "Navapur": "Nawapur",
+    "Sangole": "Sangola",
+    "Yewla": "Yevla",
+    "Jat": "Jath",
+    "Kalvan": "Kalwan",
+    "Kasba Peth": "Kasbapeth"
+}
+
+const_replacements = {
```
jdmgoogle: Why are these being replaced?

scripts/create_ocd_ids.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import print_function
+import csv
+"""
+Program to generate OCD IDs for the 2019 Indian Vidhan Sabha elections.
+"""
+
+contests = {"hr": "Haryana", "mh": "Maharashtra"}
+columns = ["id", "name"]
+country = "in"
+election = "Vidhan Sabha"
```
jdmgoogle: OCD-IDs should be independent of any one election.

scripts/create_ocd_ids.py
```diff
@@ -0,0 +1,144 @@
...
+  for d_row in districts:
+    if d_row["district"] in district_replacements:
+      district_key = district_replacements[d_row["district"]]
+    else:
+      district_key = d_row["district"]
+    ht[district_key] = d_row["abbreviation"].lower()
+
+  for c_row in consts:
+    # source of truth on district names:
+    # https://affidavit.eci.gov.in/
```
jdmgoogle: What, exactly is being pulled from there? What's the input CSV that this script is munging?

scripts/create_ocd_ids.py
rahul-nath: Will do. I think I'll use semicolons in lieu of the commas here, unless we should add new column names for constituency, district, and state (can we do that? It could potentially serve a future purpose)

scripts/create_ocd_ids.py
rahul-nath: With the electoral districts in India have changed frequently in the last decade, so there's a lot of erroneous information out there. Some of it has made its way to wikipedia, which is unfortunately the only place I'd found abbreviations. To reconcile the differences between the the ultimate source of truth (https://affidavit.eci.gov.in) and the spreadsheet of abbreviations to districts, I use this dictionary. (Actually, this particular set of replacements is going to be taken out of this PR as a more concrete source of constituencies has been found with all corresponding states and districts: https://electoralsearch.in

scripts/create_ocd_ids.py
rahul-nath: Looking at the OCD-IDs for the Lok Sabha elections, it looks like the name of the election was included in the file containing the election (I was pattern matching). I'll take this out and generalize this script better.

scripts/create_ocd_ids.py
rahul-nath: I will include another script I made that fetches data and creates constituency CSVs for each state from the new source. District and constituency information is pulled from that website. I'll detail the expected format of the district abbreviation in a comment, but that information must be retrieved from elsewhere; in this case, they were taken from wikipedia manually put into a spreadsheet without the use of any provided script.

scripts/create_ocd_ids.py
rahul-nath: Do we wish to sort by state name, then district name or state name by abbreviation?

scripts/create_ocd_ids.py
jdmgoogle: While that's useful to have, I'd prefer to split that out into a separate PR and have this one focus on only the OCD-IDs.

rahul-nath: Great. I split the PRs with this one containing the OCD IDs and [another](https://github.com/opencivicdata/ocd-division-ids/pull/177) that adds the scripts that generate them.

rahul-nath:
> Format looks good to me – I haven't checked against source of truth.

Awesome, sounds good @jpmckinney . Let me know if there's any changes that need to be done on the additional OCD-IDs

PR #176, opened 2019-10-18
ocd for 'November 2019 Romanian presidential election' and existing parliament

sguenther85:
ocd for 'November 2019 Romanian presidential election' and existing parliament


identifiers/country-ro/dispora.csv
```diff
@@ -0,0 +1 @@
+ocd-division/country:ro/region:outside_romania,Diaspora
```
jdmgoogle: Is this because of some representation (E.g., there are members of a legislative body who represent people outside the country)?

jpmckinney: It's the case in Italy (which actually has multiple diaspora seats for different regions), so I assume Romania has this, too.

jpmckinney: Also, please correct the filename (dispora.csv -> diaspora.csv).

jpmckinney: In terms of identifier, I don't think `region` is the best OCD type, because regions generally are within-country, not outside-country. Perhaps, `special:diaspora`?

PR #177, opened 2019-10-18
Add scripts that fetch district, constituency data for each Vidhan Sabha state

rahul-nath:
Scripts can be used to generalize the creation of OCD-IDs given a source-of-truth URL.


rahul-nath:
@jdmgoogle

PR #178, opened 2019-10-21
ocd for 'Jan/Feb 2020 Croatian presidential election' and existing parliament

sguenther85:
Info:
- https://en.wikipedia.org/wiki/Croatian_Parliament_electoral_districts
- https://www.sabor.hr/hr/zastupnici/izborne-jedinice


identifiers/country-hr/electoral_districts.csv
```diff
@@ -0,0 +1,13 @@
+ocd-division/country:hr,Hrvatska
+ocd-division/country:hr/ed:i~_izborna_jedinica,I. Izborna jedinica
```
jloutsenhizer: It looks like for other cases (specifically Canada) we use "ed:n" for the numbered electoral districts. What do you think about using that here?
So ocd-division/country:hr/ed:i~_izborna_jedinica would become ocd-division/country:hr/ed1

sguenther85: @jloutsenhizer look ok for me. 
But for 11 and 12 I would suggest a description that they are special electoral districts.
Some like:
 ocd-division/country:hr/ed1
 ocd-division/country:hr/ed2
.
.
ocd-division/country:hr/ed10
ocd-division/country:hr/ed11:dijaspora (for Croatian citizens living abroad)
ocd-division/country:hr/ed12:nacionalne_manjine (national minorities)

What do you think?

jloutsenhizer: Yeah I think keepign the extra text for the special districts is fine, but the colon should go after "ed"

ocd-division/country:hr/ed:1
ocd-division/country:hr/ed:2
.
.
ocd-division/country:hr/ed:10
ocd-division/country:hr/ed:11-dijaspora
ocd-division/country:hr/ed:12-nacionalne_manjine

sguenther85: Done

sguenther85:
@jpmckinney @jdmgoogle
can anybody else have a look and approve/merge if there are no problems with this PR? ;)

identifiers/country-hr/electoral_districts.csv
jpmckinney: Why the suffix for the two special districts? Their IDs seem to be 11 and 12 per https://www.sabor.hr/hr/zastupnici/izborne-jedinice

The type of division is given by the OCD type ID (`ed`). Here, we seem to be trying to communicate the type of division in the ID (`11-dijaspora`).

I think the ID should just be `11` and `12`. If you consider the type of division to be fundamentally different, then change the OCD type ID.

sguenther85: @jpmckinney 
The type of division is not "fundamentally" different in this case. My intention was to show that the two ed are "special" because they have no clear boundaries. 11 is for everyone outside croatia and 12 is for minorities all over the country.
if you still think it should be only 11 and 12, I can do that, just say ;)

jpmckinney: I think the IDs should only serve to identify the division, in which case they should be 11 and 12. Their names (after excluding the roman numeral) are different from the others, which is enough to show that they are special.

sguenther85: Okay, I'm fine with it. I changed the ids for this two entires to ":11" and ":12",

PR #180, opened 2019-10-31
update ocds for existing parliament of malta

sguenther85:


jpmckinney:
Where are these from? I found https://electoral.gov.mt/DistrettiElettorali which is quite different…

sguenther85:
@jpmckinney 
Yes and No. ;)
The list which you found is not quite different, is just more detailled.

The problem here is that the members of the parliament are elected trough the "DISTRETTI'S".
https://parlament.mt/en/13th-leg/electoral-districts/

You have here 13 electoral districts. And is this electoral districts you have local parts (or something like that).
But you have for example in the 1 DISTRETTI, 7 places included, but only 5 Members of the Parliament.
On the website above you can see which member comes from which electoral district. The members are not voted trough a single place in this DISTRETTI.

So in this case that's right, I think. 
But let me know if you are of another opinion ;)

jpmckinney:
Aha, so the elections website calls them "District 1" and the parliament's website calls them "First District" (which is what is in this PR). Looks good to me.

PR #181, opened 2019-10-31
ocds for coming "2019 United Kingdom general election" and existing parliament

sguenther85:


identifiers/country-gb/regions-eng.csv
```diff
@@ -0,0 +1,9 @@
+ocd-division/country:gb/country:gb-eng/region:ukc,North East
+ocd-division/country:gb/country:gb-eng/region:ukd,North West
+ocd-division/country:gb/country:gb-eng/region:uke,Yorkshire and the Humber
+ocd-division/country:gb/country:gb-eng/region:ukf,East Midlands
+ocd-division/country:gb/country:gb-eng/region:ukg,West Midlands
+ocd-division/country:gb/country:gb-eng/region:ukh,Eastern
+ocd-division/country:gb/country:gb-eng/region:uki,Greater London
+ocd-division/country:gb/country:gb-eng/region:ukj,South East
+ocd-division/country:gb/country:gb-eng/region:ukk,South West
```
jloutsenhizer: Are these common abbreviations for the regions? (ukc, ukd,...)

identifiers/country-gb/constituencies.csv
```diff
@@ -0,0 +1,650 @@
+ocd-division/country:gb/country:gb-eng/region:ukc/cd:berwick-upon-tweed,Berwick-upon-Tweed
```
jloutsenhizer: My only comment for this file is considered ed instead of cd for electoral district.

identifiers/country-gb/regions-eng.csv
sguenther85: Yes, these are the NUTS-Codes for the Regions.
https://en.wikipedia.org/wiki/NUTS_1_statistical_regions_of_England
http://statoids.com/ugb.html

identifiers/country-gb/constituencies.csv
sguenther85: In this case (i think) we dont have "electoral districts".
https://en.wikipedia.org/wiki/United_Kingdom_Parliament_constituencies
These are all just normal constituencies.

If we should use also ed for constituencies, then is country-de also wrong.
But please correct me if i'm wrong. Maybe I need a new clear definition of the differences between "ed" and "cd"? ;)

identifiers/country-gb/countries.csv
```diff
@@ -0,0 +1,6 @@
+id,name
+ocd-division/country:gb,United Kingdom
+ocd-division/country:gb/country:gb-eng,England
+ocd-division/country:gb/country:gb-nir,Northern Ireland
+ocd-division/country:gb/country:gb-sct,Scotland
+ocd-division/country:gb/country:gb-wls,Wales
```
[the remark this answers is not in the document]
sguenther85: Official there are no iso 3166-2 codes for these countries. they are all "gb".
For this case thay have a iso 3166-2:gb code
https://en.wikipedia.org/wiki/ISO_3166-2:GB
and there they have these codes above.

But for me it would be no problem to remove the prefix "gb-" for these countries, if you guys think so too?

identifiers/country-gb/constituencies.csv
jpmckinney: `cd` originally meant "congressional district" (for the US). However, it has been used as a more generic 'national election' district type ID, but the question of whether to continue that practice is the subject of #170 – given that many countries do not have congresses. (Similarly, `state` was used for US states, and then re-used for ADM1 divisions, though many countries do not have states.)

`ed` means "electoral district". I'm less fussed about the generic use of `ed` than `cd`, because, even if a division is locally labelled as a "constituency", "riding", "division", "ward", etc., as long as an election occurs that relates to that division, then I think it's okay (semantically) to label it as `ed`.

identifiers/country-gb/countries.csv
```diff
@@ -0,0 +1,6 @@
+id,name
+ocd-division/country:gb,United Kingdom
+ocd-division/country:gb/country:gb-eng,England
+ocd-division/country:gb/country:gb-nir,Northern Ireland
```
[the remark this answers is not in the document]
jpmckinney: 1. The `country` type ID is presently only used at the top-level. It corresponds to the concept of "sovereign country".
2. We have not yet encountered a situation in which it makes sense to have the same type ID occur at multiple levels of hierarchy within the same OCDID.

Wales, etc. are not sovereign countries, so they need a different type ID, even though they are referred to as countries locally.

NUTS1 unhelpfully puts Wales, etc. at the same level as regions of England – but NUTS is for statistical geography, whereas our project is for political geography.

I have found no good solution, but the closest I can get is:

> The Parliamentary Voting System and Constituencies Act 2011 refers to England, Scotland, Wales and Northern Ireland as "parts" of the United Kingdom in the following clause: "Each constituency shall be wholly in one of the four parts of the United Kingdom (England, Wales, Scotland and Northern Ireland)." [[source]](https://en.wikipedia.org/wiki/Countries_of_the_United_Kingdom)

So, how about `ocd-division/country:gb/part:eng,England`?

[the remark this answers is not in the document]
sguenther85: Yes, we also couldn't find a really good solution, thats the reason we tried this in the PR.
"part" is close enough for me ;)
I will change "country" to "part" for Wales, Scotland...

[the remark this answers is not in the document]
sguenther85: Done

identifiers/country-gb/constituencies.csv
sguenther85: Done

identifiers/country-gb/countries.csv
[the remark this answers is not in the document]
sguenther85: Done, "gb-" removed. And also part instead of country

identifiers/country-gb/regions-eng.csv
sguenther85: @jpmckinney any opinion here? Because you wrote above: "NUTS is for statistical geography, whereas the project is for political geography".
I didn't find any other shortcodes to describe the regions. Or should we use the long names from there regions instead of the NUTS Codes?

ocd-division/country:gb/part:eng/region:west_midlands,West Midlands ?

jloutsenhizer:
I was taking another look and realized I overlooked the OCD ID repository already contains United Kingdom OCD IDs using "uk" instead of "gb" for the country code from #156 

I'm inclined to prefer the gb country code since it's the correct ISO 3166-1 alpha-2 code for United Kingdom. Perhaps we can alias the uk OCD IDs to the new gb ones?

@jpmckinney @sguenther85 what do you think?

identifiers/country-gb/regions-eng.csv
jpmckinney: I think it's fine to re-use codes from statistical geography if they align with political geography. In this case, the statistical regions seem to have been defined based on political regions, so it's fine to use NUTS. I don't see any suitable alternatives in e.g. ISO 3166-2.

jpmckinney:
@jloutsenhizer Yes, I missed that before merging #156. We should create the aliases, as the committers for #156 are using those codes.

sguenther85:
@jloutsenhizer @jpmckinney 
I can do that, but only "country-uk.csv", and "uk_parliament_consitituencies.csv" will match to the constituencies from here.
I didn't find the regions like "ocd-division/country:gb/country:gb-eng/region:ukg,West Midlands" in the uk files. Also aliases for the "parts" are missing.

So we would have only the aliases for "gb" and all constituencies.
Should I do it anyway, even if some aliases were missing?

jloutsenhizer:
Yes, I think adding aliases only for the ones which are currently duplicated in this PR is fine. I think having duplicate OCD IDs for the same divisions is worse than having some inconsistency in the canonical OCD IDs.

sguenther85:
Ok. @jloutsenhizer @jpmckinney 
Could you take anouther look?
Have I forgotten anything?

identifiers/country-gb/regions-eng.csv
```diff
@@ -0,0 +1,9 @@
+ocd-division/country:gb/part:eng/region:ukc,North East
```
jloutsenhizer: Add csv header to this file

identifiers/country-gb/constituencies.csv
```diff
@@ -0,0 +1,650 @@
+ocd-division/country:gb/part:eng/region:ukc/ed:berwick-upon-tweed,Berwick-upon-Tweed
```
jloutsenhizer: Add csv header to this file.

sguenther85: Done

identifiers/country-gb/regions-eng.csv
sguenther85: Done

jpmckinney:
For the old OCD-IDs, @jloutsenhizer can you create a new issue tagging the participants from #156 about how we should transition to the gb OCDIDs? I figure we can either remove the old IDs (creating aliases where possible), or add a 'deprecated' column, if a hard transition isn't an option.

jpmckinney:
I re-read the OCDEP – I think perhaps we should be using "corrections" files instead of aliases. Can we remove the aliases from this PR, and move that discussion to the proposed new issue?

I also notice that the uk IDs don't have unnecessary nesting, i.e. all UK Parliament constituencies are right under country:uk, instead of nested under part and region. This is following the style in the OCDEP, e.g.:

> Cary, North Carolina (note that despite being within Wake County this is not indicated due to not being an identifying feature)
> ocd-division/country:us/state:nc/place:cary

> New York City, City Council District 36 (happens to be in Brooklyn- but not significant to include in id)
> ocd-division/country:us/state:ny/place:new_york/council_district:36

Can we remove any unnecessary hierarchy in this PR? We'd only need it if if was necessary to disambiguate two otherwise identical IDs.

jloutsenhizer:
Created #184 to continue discussion on migrating the country:uk OCD IDs.

As for these identifiers,
I find the OCDEP 2 documentation a bit vague. In particular, it's not defined what it means for something to be an "identifying feature", which is only referenced in the example identifiers. As far as I can tell this has been used for disambiguation. 

However, this leads to some strange IDs. For example, in Westmoreland County in Pennsylvania, there is a Donegal Township and a borough named Donegal. The OCD IDs respectively are:
ocd-division/country:us/state:pa/county:westmoreland/place:donegal
ocd-division/country:us/state:pa/place:donegal
It's not very clear why one has the county included as part of the OCD ID and one does not.

There's also places in Pennsylvania like Murrysville, which doesn't collide with the name of any other place in the United States, but we still include the state as a part of the OCD ID:
ocd-division/country:us/state:pa/place:murrysville

In my opinion, it makes more sense to include the full path in the OCD ID when it's part of how the place is defined. In the case of these parliament constituencies, they are defined as subdivisions of the different parts of the UK, and within England as different subdivisions of larger regions within England.

jpmckinney:
Sounds fine to me, if that is indeed how they are defined (e.g. by a boundary commission). Feel free to merge once the aliases are removed from this PR.

sguenther85:
Aliases have been removed again ;)

jpmckinney:
Thank you for your patience :)

PR #182, opened 2019-10-31
Add OCDID's for United States constituencies

ckaminer:
create new file to account for multiple state education boards
create new file for state specific constituencies
add to state courts


identifiers/country-us/us_state_courts.csv
```diff
@@ -1631,3 +1631,8 @@ ocd-division/country:us/state:nc/court_of_appeals:judge,NC Court Of Appeals Judg
 ocd-division/country:us/state:nc/district_court:11a,NC State Court - District 11A
 ocd-division/country:us/state:nc/district_court:19d,NC State Court - District 19D
 ocd-division/country:us/state:nc/superior_court:17d,NC Superior Court District 17D
+ocd-division/country:us/state:fl/court_of_appeals:1,Florida First District Court of Appeal
```
jloutsenhizer: Can you match the name formatting used for the other divisions in this file? For example:
FL State Court of Appeals - 1st District

identifiers/country-us/us-education-boards.csv
```diff
@@ -0,0 +1,59 @@
+id,name
+ocd-division/country:us/state:ks/board_of_education:1,Kansas State Board of Education District 1
```
jloutsenhizer: It looks like in other files in US we use "school_board_district" instead of "board_of_education" for the type. We should change to "school_board_district" for consistency

identifiers/country-us/state-ma-executive-offices.csv
```diff
@@ -0,0 +1,9 @@
+id,name
+ocd-division/country:us/state:ma/governors_council:1,Massachusetts Governor's Council 1st District
```
jloutsenhizer: From a quick look at wikipedia, this sounds similar to New Hampshire's Executive Council (names seem to be interchangable between the two):
https://en.wikipedia.org/wiki/Massachusetts_Governor%27s_Council
https://en.wikipedia.org/wiki/Executive_Council_of_New_Hampshire

For consitency let's use "executive_council" as the type instead of "governors_council"

ckaminer: All set 👍

identifiers/country-us/us-education-boards.csv
ckaminer: All set 👍

identifiers/country-us/us_state_courts.csv
ckaminer: All set 👍

Issue #184, opened 2019-11-04
Incorrect Country Code used for United Kingdom


jloutsenhizer:
#156  added OCD IDs for The United Kingdom but used uk for the country code rather than the "gb" from ISO-3166. #181 introduces the parliament constituencies using teh correct gb country code.

You can see #181 for some of the previous discussion. I'm proposing we use the OCD IDs from #181 and alias the IDs from #156 to these new IDs (possibly via a corrections file).

Tagging people likely interested in this change, please voice concerns if this might break how you use these identifiers.
@chris48s @showerst @symroe @sguenther85 @jpmckinney

chris48s:
Thanks for tagging me in on this.

The identifiers should definitely be migrated from `-uk` to `-gb`. I'm totally on-board with this. This is an unfortunate mistake on my part which was not caught in review. From our persepctive, it is fine for the move to be framed as corrections as opposed to aliases. I'm also happy to submit a PR migrating the code for generating the sub-national division identifiers (although I will probably defer this until after polling day for the national elections).

---

Having looked at the IDs, its unfortunate that we've applied the slugging rules to slightly different source data (unclear what the source data is for #181 ) which has lead to a couple of inconsistencies in the id slugs:

`ocd-division/country:gb/part:wls/ed:ynys_môn`
`ocd-division/country:uk/pcon:ynys_mon`

`ocd-division/country:gb/part:eng/region:uke/ed:richmond_yorks`
`ocd-division/country:uk/pcon:richmond_~yorks~`

..but this is not a big deal.

---

Unfortunately I was unable to comment before the PR was merged, but I think this next point is a big deal: I fully agree with the comments on incorrect hierarchy in https://github.com/opencivicdata/ocd-division-ids/pull/181#issuecomment-549548955 . Nesting UK Parliamentary Constituencies implies heirarchy that doesn't exist for this division type. This creates confusion when the same divison names are used for the national (UK) parliament and sub-national (Scotland, Wales, NI) parliaments/assemblies. For example, `ocd-division/country:gb/part:wls/ed:cardiff_south_and_penarth` implies the scope of the identifier is Wales, when in fact its a division of a body with UK-wide scope. In the existing `-uk` IDs, we disambiguate that using the division type e.g:

[...]

PR #186, opened 2019-11-06
migrate country:uk ids to country:gb v1

chris48s:
refs #184


jpmckinney:
Closing as #188 is where discussion was taken forward.

PR #187, opened 2019-11-11
Add ocd's for last France elections and existing parliament

sguenther85:


jdmgoogle:
Looks good. Please add a README briefly explaining what each type is and the source of the sub-country abbreviations (e.g., is there a page listing canonical abbreviations for departments).

sguenther85:
Done

identifiers/country-fr.csv
```diff
@@ -0,0 +1,699 @@
+id,name
+ocd-division/country:fr,France
+ocd-division/country:fr/oversea_department:francais_etablis_hors_de_france,Français établis hors de France
```
jpmckinney: typo: oversea -> overseas

identifiers/country-fr.csv
```diff
@@ -0,0 +1,699 @@
...
+ocd-division/country:fr/oversea_department:francais_etablis_hors_de_france/ed:1,Première circonscription des Français établis hors de France
```
jpmckinney: Didn't we start using the special id `diaspora` in other PRs? I'd prefer that consistency, instead of `francais_etablis_hors_de_france`.

sguenther85: sounds reasonable, is implemented

identifiers/country-fr.csv
sguenther85: done

jpmckinney:
@jdmgoogle @jloutsenhizer Looks good to me. I have no domain expertise for French constituencies though.

sguenther85:
Is there still any blocker here? @jdmgoogle @jloutsenhizer

jloutsenhizer:
Sorry for the delay in responding.

My only concern is if the constituency OCD IDs align with what's being discussed in #188, where we are trending toward having the hierarchy reflect the level of government which administers the constituencies.

My brief research led me to [this wikipedia article about the 2010 redistricting](https://en.wikipedia.org/wiki/2010_redistricting_of_French_legislative_constituencies#Creating_the_commission), but it's unclear to me if the Department governments are in charge of determining the districting or if the national government has some say in this.

Do you have any insight on this @sguenther85?

sguenther85:
@jloutsenhizer 
I only know that (e.g. in Germany) the national government is in charge of determining the districting. In Germany i think we have each Election a couple of new organized constituencies based on population figures.
For example it can happen that two constituencies will be merged to one or 2 constituencies get new boundaries.
I think that's what we have all over europe.

The department government will be only in charge vor local election and smaller constituencies, which are included in the one constituency which is relevant for a national election.

Does that help?

identifiers/country-fr/constituencies.csv
```diff
@@ -0,0 +1,578 @@
+id,name
+ocd-division/country:fr/region:ar/department:ai/ed:1,Première circonscription de l'Ain
```
jloutsenhizer: To keep in-line with the recent discussions, we should remove the extra nesting. So this OCD ID would become something like
ocd-division/country:fr/ed:ar-ai-1

What do you think?

sguenther85: Sorry for the long break, but you know ;) UK Elections

Actually a good idea. But of course it could be that we might run into a problem at some point, should there be 'statelists', like in Austria, Germany or Hungary, where there are national, state and constituency lists.

So I would suggest to leave them in if the construct is not fundamentally wrong?
What do you think?

jloutsenhizer: I'm not sure what you mean by "statelists", could you elaborate?

If you're talking about where different level of governments have different sets of consituencies I think we would have something like:

Country level (taking another look, I think department abbreviations are unique across all departments): ocd-division/country:fr/ed:ai-1
region-level: ocd-division/country:fr/region:ar/red:ai-1
department-level ocd-division/coutnry:fr/region:ar/district:ai/sed:1

Thoughts?

PR #190, opened 2019-12-04
Create explicit at-large US congressional districts

### LINKED ISSUE #110: Add cd:at-large to states with only one congressional district
This may not be the right place for this discussion, but...

I'd like to propose a different handling for OCD IDs for at-large districts. Currently, it seems there is no OCD ID for an at-large district, rather the next highest OCD ID division is used.

For example, Vermont has only one member of Congress, so his OCD ID is `ocd-division/country:us/state:vt`

Whereas, in a state with more than one member, they have divisions like this:

`ocd-division/country:us/state:nc/cd:2`

This makes it hard to regularize data and match between systems. It's easy to match via OCD ID, but in this case, the state level OCD ID is also shared by other elected officials (governors, senators), making matching harder and necessitating special rules for these states.

It would be better if this were handed like the Sunlight API handles them (https://sunlightlabs.github.io/congress/legislators.html), where they assign a district of 0 to at-large districts. So, for Vermont, we'd have `ocd-division/country:us/state:vt/cd:0`

jloutsenhizer:
This better aligns with the principal of having distinct OCD IDs even when the divisions happen to be coterminous. This also resolves #110


jdmgoogle:
LGTM. Merging.

PR #191, opened 2019-12-16
Add ocd's for upcoming 2020 Slovak parliamentary election - 2020-02-29

sguenther85:
The next Slovak parliamentary election will be held on 29 February 2020.


jloutsenhizer:
@jpmckinney

jpmckinney:
Noting for future reference that IDs are based on https://en.wikipedia.org/wiki/ISO_3166-2:SK

sguenther85:
> Noting for future reference that IDs are consistent with https://en.wikipedia.org/wiki/ISO_3166-2:SK

And also on HASC Code's http://www.statoids.com/usk.html ;)

PR #206, opened 2020-09-03
Add identifiers for Switzerland

1-Byte:
Add identifiers for entities in Switzerland:
 * `canton:XY`: 26 [cantons of Switzerland](https://en.wikipedia.org/wiki/Cantons_of_Switzerland)
 * `municipality:0123`: 2202 [municipalities of Switzerland](https://en.wikipedia.org/wiki/Municipalities_of_Switzerland)

In some cantons an additional level of subdivision exists. However, even if they exist they are not used for electoral purposes in all cases.

Source: https://www.bfs.admin.ch/bfs/en/home/services/geostat/swiss-federal-statistics-geodata/administrative-boundaries/generalized-boundaries-local-regional-authorities.html


identifiers/country-ch/municipalities.csv
```diff
@@ -0,0 +1,2203 @@
+id,name
+ocd-division/country:ch/canton:zh/municipality:0001,Aeugst am Albis
```
jloutsenhizer, line 2: What is the source of the munipality identifiers and are they kept stable?

If they won't necessarily be kept stable then consider using the names instead.

identifiers/country-ch/municipalities.csv
```diff
@@ -0,0 +1,2203 @@
...
+ocd-division/country:ch/canton:zh/municipality:0066,Opfikon
+ocd-division/country:ch/canton:zh/municipality:0067,Rafz
+ocd-division/country:ch/canton:zh/municipality:0068,Rorbas
+ocd-division/country:ch/canton:zh/municipality:0069,Wallisellen
+ocd-division/country:ch/canton:zh/municipality:0070,Wasterkingen
+ocd-division/country:ch/canton:zh/municipality:0071,Wil (ZH)
+ocd-division/country:ch/canton:zh/municipality:0072,Winkel
+ocd-division/country:ch/canton:zh/municipality:0081,Bachs
+ocd-division/country:ch/canton:zh/municipality:0082,Boppelsen
+ocd-division/country:ch/canton:zh/municipality:0083,Buchs (ZH)
```
jloutsenhizer, line 61: I assume the parts in parenthesis are to differentiate between municipalities of the same name in different cantons. Those shouldn't be necessary since the OCD ID uniquely identifies each municipality.

jloutsenhizer (changes requested):
Could you add a `README.md` file to this directory to document the entity types and the source of the identifiers

identifiers/country-ch/cantons.csv
```diff
@@ -0,0 +1,27 @@
+id,name
```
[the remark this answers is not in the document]
1-Byte: :+1:

identifiers/country-ch/country.csv
```diff
@@ -0,0 +1,2 @@
+id,name
+ocd-division/country:ch,Schweiz
```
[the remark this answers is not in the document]
1-Byte: Yes, this was not consistent. I added the name in all four official languages.

identifiers/country-ch/municipalities.csv
1-Byte: The number is assigned by the Federal Statistical Office and is the de facto standard to identify municipalities. The identifiers are stable and only change when municipalities merge.

identifiers/country-ch/municipalities.csv
1-Byte: The distinction would not be necessary, however I think they would still be useful as in some cases only the name will be shown to users (e.g. in a choice box).

1-Byte:
Thank for the review! I documented the sources of the identifiers in *README.md*.

identifiers/country-ch/municipalities.csv
jloutsenhizer: I think that the disambiguation strings in parentheses should be left out and we should stick with the names of the municipalities here. Whether or not those would be included in a UI would depend on that project's needs and I think it'd be easier to algorithmically add the disambiguation in than remove it.

jloutsenhizer (commented):
Sorry for the delays in followup. One last thing to resolve and then I think the PR looks good.

identifiers/country-ch/municipalities.csv
1-Byte: :+1: It is definitely easier to add them later on if needed.

1-Byte:
No need to apologize I just discovered the "Re-request review" button :blush:

1-Byte:
> Could you add a `README.md` file to this directory to document the entity types and the source of the identifiers

@jloutsenhizer Do you think it contains the necessary information or would you like me to add something?

jloutsenhizer (approved):
Apologies again for extended delays in reviewing, this looks good to me.

1-Byte:
@jloutsenhizer rebased to account for #284

jloutsenhizer:
@jpmckinney

jpmckinney (approved):

PR #209, opened 2020-09-16
Updates type for LA metro sectors IDs (la_metro_sector)

jswiesner:
There were 4 IDs introduced in https://github.com/opencivicdata/ocd-division-ids/pull/139 which contain OCD ID types (`la_metro_sector`) that are specific to one region, and therefore conceptually invalid. The OCD ID type should only provide the nature of the district, while the value should provide the specific location. 

I'd like to propose updating these IDs to instead use type `transit`. These districts appear to refer to service areas of the LA county metro system (https://media.metro.net/about_us/service_sectors/images/map_service_councils.pdf), therefore the generic type `transit` seems most suitable.

### DIFF
--- a/identifiers/country-us/state-ca-local_gov.csv
+++ b/identifiers/country-us/state-ca-local_gov.csv
@@ -100,10 +100,10 @@ ocd-division/country:us/state:ca/county:los_angeles/school_board_district:4,Los 
 ocd-division/country:us/state:ca/county:los_angeles/school_board_district:5,Los Angeles CA Unified School Board district 5
 ocd-division/country:us/state:ca/county:los_angeles/school_board_district:6,Los Angeles CA Unified School Board district 6
 ocd-division/country:us/state:ca/county:los_angeles/school_board_district:7,Los Angeles CA Unified School Board district 7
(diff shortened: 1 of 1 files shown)


jswiesner:
Hi @reginafcompton. It would be great to get your opinion about this proposed change.

reginafcompton:
Hi @jswiesner – currently, I'm not involved in the OCD project. I'd suggest pinging [Hannah Cushman](https://github.com/hancush), who supports [LA Metro Councilmatic and its scrapers](https://github.com/datamade/la-metro-councilmatic). Her insight will be more valuable than mine!

jloutsenhizer:
Using a more generic `transit` type makes sense to me, we should also add sameAs annotations to map the old OCD IDs that used `la_metro_sector` to the new ones.

Pinging @hancush to provide thoughts on this change.

hancush (changes requested):
Seems reasonable to me, @jswiesner – thanks for the PR! As @jloutsenhizer suggested, please preserve the original identifiers by adding the revised identifier to the `sameAs` field, then adding the revised identifiers as new records. [Relevant docs](http://docs.opencivicdata.org/en/latest/proposals/0002.html), if needed.

PR #202, opened 2020-04-27
Updating Zew Zealand electorates.

novitam:
Updating the general electorates for New Zealand.

Authoritative source: https://vote.nz/boundary-review-2019-2020/what-are-the-main-changes/
Summary of changes: https://vote.nz/assets/Boundary-Review/SCHEDULE-B-LIST-OF-UNCHANGED-AND-CHANGED-ELECTORATES.pdf

PR #211, opened 2020-09-24
Update NZ ValidTo and ValidFrom due to change in election dates.

novitam:
New Zealand postponed the election dates, hence the ValidFrom for new constituencies should follow.

Source: https://www.bbc.com/news/world-asia-53796434

### DIFF
--- a/identifiers/country-nz.csv
+++ b/identifiers/country-nz.csv
@@ -1,32 +1,32 @@
 id,name,validFrom,validTo
 ocd-division/country:nz,New Zealand,,
 ocd-division/country:nz/general_electorate:auckland_central,Auckland Central,,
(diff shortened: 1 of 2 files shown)


jloutsenhizer (approved):
@jpmckinney for second review

jpmckinney:
Why are some October 16? Article says October 17. (Also not sure why some used to be September 18.)

jloutsenhizer:
I believe from the last PR to update this #202 it was decided that validTo would be the day before the election and validFrom is the day of the election

jpmckinney:
Aha, I misread which column the dates were in. Thanks!

jpmckinney:
Testing @division-id-curators

PR #217, opened 2020-11-18
Add OCD-IDs for Belgium national level reporting units

ckaminer:


identifiers/country-be.csv
```diff
@@ -0,0 +1,16 @@
+id,name
+ocd-division/country:be,Belgium
+ocd-division/country:be/cd:duitstalig_kiescollege,German Language Community
+ocd-division/country:be/cd:frans_kiescollege,French Electoral College
+ocd-division/country:be/cd:nederlands_kiescollege,Dutch Electoral College
+ocd-division/country:be/province:an,Antwerp
+ocd-division/country:be/province:be,Brussels-Capital
```
azuser, line 7: I think it needs to be in a different type : region.

azuser:
It would be great to add a readme file with different type you used.

ckaminer:
> It would be great to add a readme file with different type you used.

All set 👍

identifiers/country-be/national_level.csv
```diff
@@ -0,0 +1,16 @@
+id,name
+ocd-division/country:be,Belgium
+ocd-division/country:be/region:be,Brussels-Capital
```
jdmgoogle, line 3: There also appear to be Flanders and Wallonia as the other two regions. If we're including one we should include the other two.

https://en.wikipedia.org/wiki/Communities,_regions_and_language_areas_of_Belgium#Regions

jdmgoogle (changes requested):

identifiers/country-be/national_level.csv
ckaminer: Good catch. I'm having a hard time determining what the abbreviations should be. Right now I have `region:fl` and `region:wa` but please correct me if those are wrong. I also have a related question: upon reading the Wikipedia page you linked above it looks like all of the ten provinces either belong to the Flanders or Wallonia regions. Should those IDs have the region included before the province? For example `ocd-division/country:be/region:fl/province:an` instead of `ocd-division/country:be/province:an`

PR #221, opened 2020-12-02
Add U.S. territory legislative chamber districts, GU and MP municipalities, and VI board of ed

margaret-koenig:
The territory equivalents of state legislative districts are missing, as well as municipalities in Guam and Northern Mariana Islands and the equivalent of a state board of education in the U.S. Virgin Islands.


jloutsenhizer:
Hi Margaret,

These should be added via files in the `identifiers/country-us` subdirectory. Then `scripts/compile.py` can be used to update the country-us.csv file

jloutsenhizer:
I believe #250 replaced this, closing to clean it out.

PR #222, opened 2020-12-14
Add OCD-IDs for Bulgaria national level reporting units

ckaminer:


identifiers/country-bg.csv
```diff
@@ -0,0 +1,33 @@
+id,name
+ocd-division/country:bg,Bulgaria
+ocd-division/country:bg/izbiratelni_rajoni:blagoevgrad,Blagoevgrad
```
azuser, line 3: I propose to use province as a type according to Wikipedia page : https://en.wikipedia.org/wiki/List_of_Bulgarian_constituencies.

Also, I suggest to use numbered ids : In addition to their names, constituencies are numbered from 1 to 31 according to their order in the Cyrillic alphabet. 

WDYT ? @jloutsenhizer @sguenther85

sguenther85: @azuser sounds good to me. The numbers 1-31 are stable since at least 2005.
To use province for the bulgarish translation "consticuency" sounds also good. 
To use english types names in general we should follow up.

jloutsenhizer: I'm not sure about using province for the type since the relationship between provinces and these consituencies isn't 1:1.

but I do support using the numbers of the constituencies since they appear to be used on the government website: https://www.parliament.bg/en/MP

azuser: I agree. 
`Bulgaria is divided into 28 provinces. Most of these correspond exactly to the constituencies, but Sofia City Province is divided in 3 and Plovdiv Province is divided in 2.`

Link: https://en.wikipedia.org/wiki/List_of_Bulgarian_constituencies. 

I'll update it to ed instead. WDYT ?

PR #224, opened 2020-12-16
Add national level OCD-IDs for Peru

ckaminer:


identifiers/country-pe.csv
```diff
@@ -0,0 +1,28 @@
+id,name
+ocd-division/country:pe,Peru
+ocd-division/country:pe/provincia:lp,Lima
```
azuser, line 3: We are using english as a language for other types. Is it better to use "province" instead of "provincia" ?  WDYT

PR #230, opened 2021-01-22
Add test that checks if updates have been compiled

jrunningen:
The new test can be run with Bazel, from the root of the repository:

```
$ bazel test :all --test_output=errors
```

Any change to this repository that modifies sources for some country's data (the stuff in `identifiers/*/*.csv`) should also modify the corresponding compiled output (`identifiers/country-??.csv`), such that they match.

This test runs the compiler for every country, and compares its output to what's been committed, ensuring they are equal. If this test fails, you need to re-run the compiler for your country, and commit the result.

The test currently fails for countries `mt`, `co`, and `au`. We can follow up with fixes to the data to make the test pass.

Sample output of the test can be seen at https://gist.github.com/jrunningen/4e59a37f83bcf099011e63029176ac8a

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,8 @@
+py_test(
+    name = "compile_test",
+    srcs = ["test/compile_test.py"],
+    data = glob([
+        "identifiers/**/*.csv",
+    ]) + ["scripts/compile.py"],
+    main = "test/compile_test.py",
+)
--- a/test/compile_test.py
+++ b/test/compile_test.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+# Run this test from the root of the repository, as:
+# $ bazel test :all --test_output=errors
+
+import difflib
+import os
+import re
+import subprocess
+import unittest
+
+
+class TestSourcesMatchCommittedCSV(unittest.TestCase):
+  """Check that what's committed matches the output of the compiler.
+
+  Any change to this repository that modifies sources for some country's data
+  (the stuff in identifiers/*/*.csv) should also modify the corresponding
+  compiled output (identifiers/country-??.csv), such that they match.
+
+  This test runs the compiler, and compares its output to what's been
+  committed, ensuring they are equal. If this test fails, you need to re-run
+  the compiler for your country, and commit the result.
+  """
+
+  def get_committed_and_compiled_data(self, country_code):
+    # Read what's in the repo.
+    committed_csv_path = F'identifiers/country-{country_code}.csv'
+    try:
+      with open(committed_csv_path, 'r') as committed_file:
+        committed_csv = committed_file.read()
+    except FileNotFoundError:
+      # In the case that the commited csv file does not exist, we treat it as an
+      # empty string.
+      committed_csv = ''
+
+    # Run the compiler with temp output.
+    compiler_output_path = os.path.join(os.getenv('TEST_TMPDIR'), F"out-{country_code}.csv")
+    subprocess.run([
+        './scripts/compile.py', F'--output_csv={compiler_output_path}', country_code
+    ], check=True)
+
+    # Read the output of the compiler.
+    with open(compiler_output_path, 'r') as compiler_output_file:
+      compiler_output = compiler_output_file.read()
+
+    return committed_csv, compiler_output
+
+  def test_all_countries(self):
+    _, csv_source_dirs, _ = list(os.walk('identifiers'))[0]
+    mismatches = []
+    for dirname in csv_source_dirs:
+      # Ignore any files that don't match the expected pattern.
+      m = re.match(r'country-(..)', dirname)
+      if m:
+        country_code = m.group(1)
+        committed_csv, compiler_output = self.get_committed_and_compiled_data(
+            country_code)
+        if committed_csv != compiler_output:
+          mismatches.append(country_code)
+    # Rather than assert as we go, which would bail out on the first failure, we
+    # compile a list of all the mismatching countries, and print a message that
+    # summarizes what to do.
+    error_msg = ('The following countries must be recompiled to match their '
+                 'source files:\n  ' + '\n  '.join(mismatches) +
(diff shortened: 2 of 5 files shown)


BUILD
```diff
@@ -0,0 +1,8 @@
+py_test(
```
jloutsenhizer, line 1: I'm not sure what the best practice here is, but do you think this BUILD file should be in the test subdir?

test/compile_test.py
```diff
@@ -0,0 +1,66 @@
...
+      compiler_output = compiler_output_file.read()
+
+    return committed_csv, compiler_output
+
+  def test_all_countries(self):
+    _, _, master_csv_files = list(os.walk('identifiers'))[0]
+    mismatches = []
+    for filename in master_csv_files:
+      # Ignore any files that don't match the expected pattern.
+      m = re.match(r'country-(..)\.csv', filename)
```
jloutsenhizer, line 48: I think this will only catch countries with existing country-*.csv. If a new country is added without being compiled, then we would miss it. Could you update this to take into account ones which don't have a csv file compiled already?

test/compile_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+  def get_committed_and_compiled_data(self, country_code):
+    # Read what's in the repo.
+    committed_csv_path = F'identifiers/country-{country_code}.csv'
+    with open(committed_csv_path, 'r') as committed_file:
+      committed_csv = committed_file.read()
+
+    # Run the compiler with temp output.
+    compiler_output_path = os.path.join(os.getenv('TEST_TMPDIR'), F"out-{country_code}.csv")
+    subprocess.run([
```
jloutsenhizer, line 33: I haven't tried this out, but what happens if the directory of country specific files hasn't been created. Does the compiler fail and the test fail to run? Or do we still get a sane output?

I think ideally we'd be able to just tell the user that there's a mismatch

jloutsenhizer (changes requested):
I also see an empty WORKSPACE file created in this PR. If that's not needed for bazel it should be removed.

BUILD
jrunningen: I think it makes sense conventionally to be at the root, similar to a Makefile. You'd typically run your make/bazel commands from the root of the repo.

Also, `BUILD` references files in `scripts` and `identifiers`, too, so to access those, they need to be underneath a directory with a `BUILD` file. So, if there's no top-level `BUILD`, we'd need a separate `BUILD` in each subdir. It seems cleaner this way.

test/compile_test.py
jrunningen: Currently, if `identifiers/country-xx.csv` exists, but `identifiers/country-xx/` does not:

- The test fails, asking you to run `./scripts/compile.py xx`
- Running that command will overwrite `identifiers/country-xx.csv` with a blank file, which is probably not what we want.

I'll see if I can update the test to ignore country xx in this case.

test/compile_test.py
jrunningen: Currently, if identifiers/country-yy.csv does not exist, but identifiers/country-yy/foo.csv does, the test will ignore yy.

I'll see if I can update the test to prompt the user to run the compiler in that case.

test/compile_test.py
jrunningen: Done.

test/compile_test.py
jrunningen: Done.

jrunningen:
The WORKSPACE file, although empty, is required by Bazel. See https://docs.bazel.build/versions/master/build-ref.html#workspace.

BUILD
jloutsenhizer: Good point about the different directory access and having it be used in a more similar style to a Makefile.

jloutsenhizer (approved):
This looks good to me and useful to have since currently we have to manually check if the `country-xx.csv` file looks like it was properly compiled.

@jpmckinney I added you as a second reviewer, but I'm not sure if anyone else should also be looped in since this involves adding a new tool dependency for the project on bazel

jpmckinney (approved):
I've never used Bazel. Is it really doing a lot more than this?

`python -m unittest test/compile_test.py`

jrunningen:
Bazel has a few advantages over running the test directly.

- `bazel test :all` will run all the tests defined in BUILD, so if we add more tests, in different files (or different languages, even), we don't have to change the testing procedure for running all the tests. In a future pull request, I was hoping to add a [Github action that invokes the tests this way](https://github.com/jrunningen/ocd-division-ids/blob/ci/.github/workflows/test.yml) so that invalid pull requests can be caught before merge.
- It provides an isolated directory (`TEST_TMPDIR`) where the test can write out compiled files for comparison to the original. The test can write files without having to do its own cleanup afterward, and it definitely won't mess with the working tree.

jpmckinney:
> Bazel has a few advantages over running the test directly.
> 
> * `bazel test :all` will run all the tests defined in BUILD, so if we add more tests, in different files (or different languages, even), we don't have to change the testing procedure for running all the tests. In a future pull request, I was hoping to add a [Github action that invokes the tests this way](https://github.com/jrunningen/ocd-division-ids/blob/ci/.github/workflows/test.yml) so that invalid pull requests can be caught before merge.
> * It provides an isolated directory (`TEST_TMPDIR`) where the test can write out compiled files for comparison to the original. The test can write files without having to do its own cleanup afterward, and it definitely won't mess with the working tree.

You can call `python -m unittest` to just let it discovery test files, and the test code could use the tempfile module from the standard library instead. I really hope we don’t write tests in other languages :) Anyway, it’s just one dependency, so no big deal.

jrunningen:
Ah, good point. I may have just reached for the thing I'm used to. If you like, I'm happy to remove the dependency on Bazel. Just let me know.

jpmckinney:
All good, I think! I don't have the time to change the code to use the tempfile module, and it's just one dependency. We can always refactor later if there's ever an issue.

PR #231, opened 2021-01-26
Update au, co, and mt to match source files

jrunningen:
These countries did not match their sources. I've re-run the compiler for each one. This was caught by the `bazel test compile_test` (output at https://gist.github.com/jrunningen/4e59a37f83bcf099011e63029176ac8a).

This was found with the test that was added in #230.


jloutsenhizer (approved):
Thanks for cleaning up these errors

jpmckinney:
Thank you!

PR #234, opened 2021-01-28
Populate country level OCD IDs for all countries in the world

jrunningen:
For any country that did not yet have a CSV under identifiers, this
commit adds a CSV that defines the country code for that country.

This was created using the list of officially assigned code elements
from Wikipedia at
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements,
at the time of this commit.

Any countries that already existed in this repository were not touched.

### DIFF
--- a/identifiers/country-aq.csv
+++ b/identifiers/country-aq.csv
@@ -1,2 +1,0 @@
-id,name
-ocd-division/country:aq,Antarctica
--- a/identifiers/country-aq.csv
+++ b/identifiers/country-aq.csv
@@ -1,0 +1,2 @@
+id,name
+ocd-division/country:aq,Antarctica
--- a/identifiers/country-as.csv
+++ b/identifiers/country-as.csv
@@ -1,0 +1,2 @@
+id,name
+ocd-division/country:as,American Samoa
--- a/identifiers/country-as.csv
+++ b/identifiers/country-as.csv
@@ -1,2 +1,0 @@
-id,name
-ocd-division/country:as,American Samoa
--- a/identifiers/country-bm.csv
+++ b/identifiers/country-bm.csv
@@ -1,0 +1,2 @@
+id,name
+ocd-division/country:bm,Bermuda
--- a/identifiers/country-bm.csv
+++ b/identifiers/country-bm.csv
@@ -1,2 +1,0 @@
-id,name
-ocd-division/country:bm,Bermuda
--- a/identifiers/country-gu.csv
+++ b/identifiers/country-gu.csv
@@ -1,0 +1,2 @@
+id,name
+ocd-division/country:gu,Guam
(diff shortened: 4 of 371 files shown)


identifiers/country-aq.csv
```diff
@@ -0,0 +1,2 @@
+id,name
+ocd-division/country:aq,Antarctica
```
jloutsenhizer, line 2: I don't think this OCD ID would be correct since Antarctica isn't a country.

identifiers/country-as.csv
```diff
@@ -0,0 +1,2 @@
+id,name
+ocd-division/country:as,American Samoa
```
jloutsenhizer, line 2: American Samoa is also not a country, but a territory of the US

identifiers/country-gu.csv
```diff
@@ -0,0 +1,2 @@
+id,name
+ocd-division/country:gu,Guam
```
jloutsenhizer, line 2: Same as American Samoa

identifiers/country-bm.csv
```diff
@@ -0,0 +1,2 @@
+id,name
+ocd-division/country:bm,Bermuda
```
jloutsenhizer, line 2: This is a territory of the UK, so I don't think it should be a separate country.

jloutsenhizer (changes requested):
There's ISO codes for a lot of entities which aren't independent countries. This wiki page has a useful table: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

I think we should probably filter this down to limit to those that are actually countries.

Also it looks like this PR directly creates this `identifiers/country-*.csv` files, but I think we would want the subdirs for each of these to be created as well so that they can each be compiled with the script.

jrunningen:
All comments resolved. See the commit messages for details.

jrunningen:
For the failing test, see #237.

jrunningen:
I've rebased atop the latest merge to fix the test.

jloutsenhizer (approved):
I think this looks good to me based on spot checking the added OCD IDs.

jpmckinney:
To clarify, how were the non-countries filtered out? Are the countries based on the table here, where Sovereignty is either "disputed" or "UN member state"? https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

jrunningen:
Yes, it's based on [that table](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes#Current_ISO_3166_country_codes). The full commit message in eb5e2789eef8d51fa02cc0ad0d4bf981f423762a describes the criteria for filtering a non-country:

```
This is any country code with sovereignty not listed as "UN member
state," "UN observer state," or "disputed," and which did not previously
exist in the respository.
```

jpmckinney (approved):

PR #235, opened 2021-01-28
Add Github action to run tests for pushes and pull requests

jrunningen:
This adds a github action that will invoke `bazel test :all`. Sample
results can be seen here:

- Passing: https://github.com/jrunningen/ocd-division-ids/actions/runs/513136468
- Failing: https://github.com/jrunningen/ocd-division-ids/actions/runs/513132397

This action will run for new pull requests, so the reveiwer is notified
of problems like the inputs not compiling, or the top-level country CSV
not being updated to match sources.

`compile.py` is also updated here to enforce UTF-8 encoding, which otherwise can differ by locale.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,16 @@
+name: Test
+
+# We want testing for new pull requests before they are merged.
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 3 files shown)


jloutsenhizer (approved):

jpmckinney:
Oops, I assumed the tests had been run on the repo, not just the fork. Someone will need to recompile ec (Ecuador): https://github.com/opencivicdata/ocd-division-ids/runs/1818393472?check_suite_focus=true

PR #238, opened 2021-02-17
Add India Vidhan Parishad OCD-ID

novitam:


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-in/README
```diff
@@ -1,11 +1,14 @@
+
```
jloutsenhizer, line 1: nit: extra newline added

identifiers/country-in/README
```diff
@@ -1,11 +1,14 @@
...
 README
```
jloutsenhizer, line 2: Could you add to the README a descritpion of the types being used within the `country-in` dir and what types of districts they map to?

identifiers/country-in/vidhan_parishad.csv
```diff
@@ -0,0 +1,215 @@
+id,name
+ocd-division/country:in/state:ap/special_constituency:graduate,Graduate special constituency
```
jloutsenhizer, line 2: I think `special` should be used instead of `special_constituency` for a bit more consistency with other country OCD IDs.

identifiers/country-in/vidhan_parishad.csv
```diff
@@ -0,0 +1,215 @@
...
+ocd-division/country:in/state:ap/special_constituency:teacher/ed:guntur_krishna,Guntur-Krishna
+ocd-division/country:in/state:ap/special_constituency:teacher/ed:srikakulam_vizianagaram_vishakapatnam,Srikakulam-Vizianagaram-Vishakapatnam
+ocd-division/country:in/state:br/special_constituency:graduate,Graduate special constituency
+ocd-division/country:in/state:br/special_constituency:graduate/ed:darbhanga,Darbhanga
+ocd-division/country:in/state:br/special_constituency:graduate/ed:gaya,Gaya
+ocd-division/country:in/state:br/special_constituency:graduate/ed:koshi,Koshi
+ocd-division/country:in/state:br/special_constituency:graduate/ed:patna,Patna
+ocd-division/country:in/state:br/special_constituency:graduate/ed:saran,Saran
+ocd-division/country:in/state:br/special_constituency:graduate/ed:tirhut,Tirhut
+ocd-division/country:in/state:br/special_constituency:local_body,Local body special constituency
```
jloutsenhizer, line 34: Shouldn't this be `local_authority` like the others?

jloutsenhizer (changes requested):
This is an interesting case where districts exist for different groups of people crossed with different geographical distircts.

What we have for `EC` is the closest example I see to this where we have `special:diaspora` for overseas residents and then an `ed` OCD ID hanging off of that for those groups sliced by geographical regions.

The only alternative I see would be to define a special type for each type of electoral district type like `graduates_ed` etc, but I think I like this structure for the consistency with `EC`. Tagging @jpmckinney early in case he has some thoughts on this structure.

jpmckinney:
Hmm, can you share some links for reference? I'm not sure how to interpret graduate, teacher, etc.

jpmckinney:
Aha: https://en.wikipedia.org/wiki/State_legislative_councils_of_India

Could we split the .../special_constituency:.../ed:... OCD-IDs into a separate file? It will be easier to review the .../special_constituency:... level.

identifiers/country-in/README
novitam: Added, PTAL

identifiers/country-in/vidhan_parishad.csv
novitam: Bihar uses `local_body` to refer to this division instead, and so we are following the state / ECI to be consistent.

identifiers/country-in/vidhan_parishad.csv
novitam: Changed to `special`

novitam:
@jpmckinney Background on Vidhan Parishad (teachers, graduate, etc):

 The members of the Legislative Council are elected as well as nominated. Broadly speaking, 5/6th of the total members of the Council are indirectly elected and 1/6th is nominated by the Governor. The composition is as follows:
a) A 1/3 of the total members of the Council are elected by electorates consisting of members of local bodies such as city corporations, municipalities and Zila Parishads.
b) Another 1/3 are elected by members of the Legislative Assembly from among the persons who are not already members of the Legislative Assembly.
c)  1/12 are elected by electorates consisting of persons who are graduates of three years standing, residing in that state.
d)  1/12 are elected by electorates consisting of persons engaged for at least three years in teaching in educational institutions within the state, not lower in standard than secondary schools.
e)  The remainder 1/6 are nominated by the Governor from persons having knowledge or practical experience in fields such as legislature, science, arts, co-operative movement and social service.

Link: https://en.wikipedia.org/wiki/State_legislative_councils_of_India

In this pull request:
* special:teacher maps to part (d) above.
* special:graduate maps to part (c) above.
* special:local body / local authority maps to part (a) above.
* no ocd-ids created for (b) because the state level OCD-ID could represent the MLA for that state
* no ocd-ids created for (e) because these are by nomination / appointment, and does not require them being elected

novitam:
@jpmckinney done with the change you requested.

identifiers/country-in/README
```diff
@@ -1,11 +1,27 @@
...
\ No newline at end of file
+*    Vidhan Sabha (except for Jammu & Kashmir)
+*    Vidhan Parishad
+
+OCD ID Types used in `country-in`:
+*    state - represents the states in India
+*    territory - represents the union territories of India
+*    district - electoral districts delimited by the Election Commissions of India
+*    special:graduate - electoral division only for graduates used 
+*    special:teacher - electoral division only for graduates used
```
jloutsenhizer, line 22: graduates -> teachers

identifiers/country-in/vidhan_parishad.csv
jloutsenhizer: Thanks for clarifying, consistency with the local naming sounds good to me.

identifiers/country-in/special.csv
```diff
@@ -0,0 +1,19 @@
+id,name
```
jloutsenhizer, line 1: I think naming this file `vidhan_parishad_special.csv` or `vidhan_parishad_special_constituencies.csv` would make it clearer this is specifically for Vidhan Parishad

identifiers/country-in/README
novitam: Good catch :)

identifiers/country-in/special.csv
novitam: Done.

jloutsenhizer (approved):

novitam:
@jpmckinney please review for 2nd approval

novitam:
Alternatively, @sguenther85 please review for 2nd approval. Thank you!

jpmckinney (approved):

PR #239, opened 2021-02-26
Add Ethiopia OCD ID's

4bic:


4bic:
Hi @jpmckinney, whenever you can, do review,  we'd like to add a few countries we are looking at.

identifiers/country-et/states.csv
```diff
@@ -0,0 +1,13 @@
+id,name
+ocd-division/country:et,Ethiopia
+ocd-division/country:et/regional_state:af,Afar
```
jpmckinney, line 3: Why `regional_state` and not `state`?

identifiers/country-et/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This folder contains the following Ethiopian administrative divisions :
+
+## State
+* Regional states, represented as *state* and city administrations represented as *administration*
+
+The OCD identifiers are created based on the ISO 3166-2 code [(source)](https://en.wikipedia.org/wiki/ISO_3166-2:ET)
+
+## Constituencies
+* Constituencies represented as *ed*, soured from [hopr.gov.et](http://hopr.gov.et/)
```
jpmckinney, line 11:
```suggestion
* Constituencies represented as *ed*, sourced from [hopr.gov.et](http://hopr.gov.et/)
```

identifiers/country-et/constituencies.csv
```diff
@@ -0,0 +1,760 @@
+id,name,sameAs,sameAsNote
+ocd-division/country:et,Ethiopia,,
+ocd-division/country:et/administration:aa,Addis Ababa,,
+ocd-division/country:et/administration:aa/ed:44075,44075,,
+ocd-division/country:et/administration:aa/ed:werda_1_and_9,Werda 1/9,ocd-division/country:et/administration:aa/ed:woreda_1_and_9,Possible spelling change across terms
```
jpmckinney, line 5: Are the alternative spellings necessary? We typically only create "sameAs" entries when renaming an OCD-ID – not when creating it.

jpmckinney (commented):
I left a few comments. I'm assuming the total constituencies matches the official source.

@jloutsenhizer For additional review.

identifiers/country-et/states.csv
4bic: @jpmckinney , using `state` definitely is logical in this instance, our dilemma was that almost all official references [like this one](http://www.mfa.gov.et/Home/TemplateI?PageName=Government&TYPE=About%20Ethiopia&SUBTYPE=About%20Ethiopia&Language=English)  refer to these admin regions as  `regional states`  cc @jloutsenhizer

identifiers/country-et/constituencies.csv
4bic: Got you, thank you for that, so something that may need more clarification would be how to capture
instances where a region has multiple versions of the name eg [ Mekelle / Mekele / Mekale ](https://en.wikipedia.org/wiki/Mekelle)

jpmckinney: I think in this project we just adopt one name. There are many projects, like GeoNames, which track alternative names, but that is not the goal of this project.

identifiers/country-et/states.csv
jpmckinney: I can go either way - I'd like the opinion of a second reviewer.

identifiers/country-et/constituencies.csv
4bic: Alright, this will be cleaned up and updated.

identifiers/country-et/states.csv
jloutsenhizer: My preference would be for `state` rather than `regional_state` since it'd reduce the number of unique identifier types for first-level adminstrative areas.

jloutsenhizer (changes requested):
Overall I think this is in good shape once `state` vs `regional_state` is resolved and the `sameAs` annotations for the spellings of the electoral districts is cleaned up

identifiers/country-et/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This folder contains the following Ethiopian administrative divisions :
+
+## State
+* Regional states, represented as *state* and city administrations represented as *administration*
+
+The OCD identifiers are created based on the ISO 3166-2 code [(source)](https://en.wikipedia.org/wiki/ISO_3166-2:ET)
+
+## Constituencies
+* Constituencies represented as *ed*, sourced from [hopr.gov.et](http://hopr.gov.et/)
```
jloutsenhizer, line 11: Is there a more specific page to link to for the source of the constituencies?  The website is down currently so browsing through archive.org is slow, but I'm having trouble finding the correct page.

identifiers/country-et/states.csv
```diff
@@ -0,0 +1,13 @@
+id,name
+ocd-division/country:et/state:af,Afar Region
```
jloutsenhizer, line 2: nit: can you sort this file?

identifiers/country-et/constituencies.csv
```diff
@@ -0,0 +1,653 @@
+id,name,sameAs,sameAsNote
+ocd-division/country:et/administration:aa/ed:44075,44075,,
```
jloutsenhizer, line 2: It looks like there's too many constituencies. I count 652 in this file, but my understanding is that there's only 547 seats: https://en.wikipedia.org/wiki/House_of_Peoples%27_Representatives

jloutsenhizer (changes requested):

PR #242, opened 2021-03-04
Add OCD-ID's of Cyprus

sguenther85:


sguenther85:
Hi. @jpmckinney @azuser @jloutsenhizer

Can somebody please review?
Ocd's are for upcoming election on 2021-05-30.

Thanks! ;)

identifiers/country-cy/top_level.csv
```diff
@@ -1,2 +1,8 @@
 id,name
 ocd-division/country:cy,Cyprus
+ocd-division/country:cy/district:01,Lefkosia
+ocd-division/country:cy/district:02,Lemesos
+ocd-division/country:cy/district:03,Larnaka
+ocd-division/country:cy/district:04,Ammochostos
+ocd-division/country:cy/district:05,Pafos
+ocd-division/country:cy/district:06,Keryneia
```
jpmckinney, line 8: I suggest omitting the initial 0.

jpmckinney: Also, where do these numbers come from?

identifiers/country-cy/top_level.csv
```diff
@@ -1,2 +1,8 @@
 id,name
 ocd-division/country:cy,Cyprus
+ocd-division/country:cy/district:01,Lefkosia
```
azuser, line 3: I would use **ed** instead of **district**. WDYT ?

identifiers/country-cy/top_level.csv
sguenther85: These are the official ISO 3166-2:CY of these districts. Also with the leading 0. You can find the sources in the Readme.
https://en.wikipedia.org/wiki/ISO_3166-2:CY

Should we still omit the leading 0, although they are specifically listed as ISO 3166-2?

identifiers/country-cy/top_level.csv
sguenther85: I was going to do that at first, but it's really districts (https://en.wikipedia.org/wiki/Districts_of_Cyprus) and not electoral districts like in ecuador (https://es.wikipedia.org/wiki/Distritos_electorales_del_Ecuador).

But if everyone else thinks ed would be better anyway, i don't have a problem with that either? does anyone else have an opinion on this?

identifiers/country-cy/top_level.csv
jpmckinney: All good - I missed the second link.

identifiers/country-cy/top_level.csv
jpmckinney: I think district is fine/correct in this case

jpmckinney (approved):

identifiers/country-cy/top_level.csv
azuser: Thanks for the clarification. Makes sense.

jloutsenhizer: `district` sounds good to me as well in this use-case

jloutsenhizer (approved):

PR #244, opened 2021-03-09
Make the test ignore row ordering

jrunningen:
Row ordering shouldn't be significant in this comparison. The compile.py
script will sort the rows before writing the output, which can be
surprising to someone who adds just one source file, copies it exactly
to identifiers/country-??.csv, but gets hit with a test failure because
the rows aren't sorted.

This change makes the test pass even if the rows aren't sorted.

### DIFF
--- a/test/compile_test.py
+++ b/test/compile_test.py
@@ -10,6 +10,16 @@ import subprocess
 import unittest
 
 
+def sort_csv_lines(csv_text):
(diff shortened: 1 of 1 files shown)


jrunningen:
@jloutsenhizer

jloutsenhizer (approved):

PR #247, opened 2021-03-12
Add OCD-ID's of Czech Republic

sguenther85:


sguenther85:
Hi,

@jrunningen the compiler throws an error becaus of letters in the identifier.
`←[91mERROR: validation error in identifiers/country-cz/districts.csv: invalid id: ocd-division/country:cz/region:20/district:20A ←[0m
`
Maybe you can take a look and maybe also fix the issue (if its a issue) in the compiler?

@jloutsenhizer @jpmckinney @azuser 
The content should be fine i think. You can find all important links for the identifiers in the Readme.

Let me know if anything looks incorrect.

Best
Steve

sguenther85:
Sry guys, found an issue according to districts to regions. Have to take a second look. I was too early this time. ;)

PR #257, opened 2021-06-08
Add ocd-ids for South Korea National Assembly

novitam:
South Korea National Assembly is a unicameral parliament, with 253 members elected from single-member constituencies (https://en.wikipedia.org/wiki/South_Korean_Legislature_Constituencies), and 47 from nationwide whereby 30 are using additional member system and the remaining 17 are via proportional representation.

3 types of special cities are used according to the South Korea administrative divisions: https://en.wikipedia.org/wiki/List_of_special_cities_of_South_Korea#Position_in_hierarchy_and_types

### DIFF
--- a/identifiers/country-kr/cities.csv
+++ b/identifiers/country-kr/cities.csv
@@ -1,9 +1,9 @@
 id,name
 ocd-division/country:kr/metropolitan_city:26,Busan
 ocd-division/country:kr/metropolitan_city:27,Daegu
-ocd-division/country:kr/metropolitan_city:28,Incheon
-ocd-division/country:kr/metropolitan_city:29,Gwangju
 ocd-division/country:kr/metropolitan_city:30,Daejeon
+ocd-division/country:kr/metropolitan_city:29,Gwangju
+ocd-division/country:kr/metropolitan_city:28,Incheon
 ocd-division/country:kr/special_self_governing_city:50,Sejong
 ocd-division/country:kr/special_city:11,Seoul
 ocd-division/country:kr/metropolitan_city:31,Ulsan
--- a/identifiers/country-kr/cities.csv
+++ b/identifiers/country-kr/cities.csv
@@ -1,9 +1,9 @@
 id,name
 ocd-division/country:kr/metropolitan_city:26,Busan
 ocd-division/country:kr/metropolitan_city:27,Daegu
-ocd-division/country:kr/metropolitan_city:30,Daejeon
-ocd-division/country:kr/metropolitan_city:29,Gwangju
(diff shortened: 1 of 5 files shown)


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-kr/cities.csv
```diff
@@ -0,0 +1,9 @@
+id,name
```
jloutsenhizer, line 1: Please sort this file

identifiers/country-kr/single_member_constituencies.csv
```diff
@@ -0,0 +1,253 @@
+id,name
```
jloutsenhizer, line 1: It looks like there's only 252 entries in this file, but the number of single-member districts is supposed to be 253, so one must be missing.

identifiers/country-kr/single_member_constituencies.csv
```diff
@@ -0,0 +1,253 @@
+id,name
+ocd-division/country:kr/metropolitan_city:26/ed:central_district_and_yeong_island,Central District and Yeong Island
```
jloutsenhizer, line 2: From https://en.wikipedia.org/wiki/South_Korean_Legislature_Constituencies

It sounds like the National Eleciton Commission handles the determination of the districts for South Korea's Legislature. Given that, I think the districts should hang off of country:kr instead of off of the individual admin area 1 OCD IDs.

jloutsenhizer (changes requested):

identifiers/country-kr/cities.csv
novitam: It was sorted by `id`. I have now sorted it by `name`.

identifiers/country-kr/single_member_constituencies.csv
novitam: Thanks! Found the missing one.

identifiers/country-kr/single_member_constituencies.csv
novitam: Done.

identifiers/country-kr/cities.csv
jloutsenhizer: I think `id` is the correct column to sort by, but `ocd-division/country:kr/metropolitan_city:31` was at the end out of order. Could you resort by `id`?

novitam: Done

novitam:
@jpmckinney please review as second approver, thank you!

jpmckinney (approved):
I haven't checked against ground truth, but structurally looks good!

PR #259, opened 2021-06-11
Add 1 missing constituency in AP + aliases to 3 constituencies in Vidhan Parishad, India.

novitam:
Current ocd-ids list contain the list of constituencies in the last delimitation exercise. However, the list of members / office holders by the government refers to an updated constituency names (although there isn't any change in the constituency itself). Such renaming exercise is common in India.

For example, constituency Thane as was listed as Thane in the delimitation exercise, but the the officeholder for that constituency is listed as officeholder for Thane - Phalghar.

Delimitation Document: https://legislative.gov.in/sites/default/files/%2811%29%20THE%20DELIMITATION%20OF%20COUNCIL%20CONSTITUENCIES%202%5B%28MAHARASHTRA%29%5D%20ORDER%2C%201951_0.pdf

Member list by the government:
http://mls.org.in/pdf2020/winter/28%20December%202020.pdf

Constituencies list:
https://en.wikipedia.org/wiki/Maharashtra_Legislative_Council


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-in/aliases.csv
```diff
@@ -1,546 +1,549 @@
-id,sameAs,sameAsNote
```
jloutsenhizer, line 1: Any idea what's going on with this file? It looks like just the last 3 lines were added. Ideally the diff would just be that so that the version history is easier to browse.

jloutsenhizer (changes requested):
Consider also updating the title to reflect we're adding the Krishna which was misisng from the Andhra Pradesh districts.

identifiers/country-in/aliases.csv
novitam: No idea, I copied the raw text from the original file vs the modified one into a diff tool, and only the last 3 lines are highlighted as the diff.
Screenshot of diff

jloutsenhizer: Strange, my only guess is that maybe the line terminators changed, one might have had `CR LF` and the other just `LF`. Either way I don't think it's necessarily blocking.

jloutsenhizer (approved):

novitam:
@jpmckinney please review and approve, thanks!

jpmckinney:
@novitam In terms of "ground truth", is the only change that the division changed name? If so, we don't need to mint new IDs. We can just change the name. This does mean that the name and ID no longer 100% match, but that is generally the case with OCDIDs (e.g. some IDs are numeric).

novitam:
As far as I understand, yes only the name has changed. In the next delimitation exercise, the boundaries would reflect the new name, but it is unclear if the constituency will also be changed along with the name. Please let me know how to proceed for this case. Thanks!

jpmckinney:
Hmm, I'm not sure what the best approach is here. The current alias is safe – worst case, we just created extra IDs. So, I'll merge as-is.

novitam:
Thank you!

PR #262, opened 2021-07-08
Add ocd-ids for Malaysia Dewan Negara and Dewan Rakyat.

novitam:
Malaysia has bicameral parliament:

1. Dewan Negara (Senate)

-  26 members elected by state legislative assembly from 13 constituencies, where each constituencies elect 2 members.
- https://en.wikipedia.org/wiki/State_legislative_assemblies_of_Malaysia

2. Dewan Rakyat (House of Representatives)

- 222 members elected by single member constituencies.
- https://en.wikipedia.org/wiki/List_of_Malaysian_electoral_districts

ISO 3166-2:MY: https://en.wikipedia.org/wiki/ISO_3166-2:MY

### DIFF
--- a/identifiers/country-my/constituencies.csv
+++ b/identifiers/country-my/constituencies.csv
@@ -112,18 +112,18 @@ ocd-division/country:my/state:10/ed:p110,Klang
 ocd-division/country:my/state:10/ed:p111,Kota Raja
 ocd-division/country:my/state:10/ed:p112,Kuala Langat
 ocd-division/country:my/state:10/ed:p113,Sepang
-ocd-division/country:my/federal_territory:14/ed:p114,Kepong
-ocd-division/country:my/federal_territory:14/ed:p115,Batu
-ocd-division/country:my/federal_territory:14/ed:p116,Wangsa Maju
-ocd-division/country:my/federal_territory:14/ed:p117,Segambut
-ocd-division/country:my/federal_territory:14/ed:p118,Setiawangsa
-ocd-division/country:my/federal_territory:14/ed:p119,Titiwangsa
-ocd-division/country:my/federal_territory:14/ed:p120,Bukit Bintang
-ocd-division/country:my/federal_territory:14/ed:p121,Lembah Pantai
-ocd-division/country:my/federal_territory:14/ed:p122,Seputeh
-ocd-division/country:my/federal_territory:14/ed:p123,Cheras
-ocd-division/country:my/federal_territory:14/ed:p124,Bandar Tun Razak
-ocd-division/country:my/federal_territory:16/ed:p125,Putrajaya
+ocd-division/country:my/territory:14/ed:p114,Kepong
+ocd-division/country:my/territory:14/ed:p115,Batu
+ocd-division/country:my/territory:14/ed:p116,Wangsa Maju
+ocd-division/country:my/territory:14/ed:p117,Segambut
+ocd-division/country:my/territory:14/ed:p118,Setiawangsa
+ocd-division/country:my/territory:14/ed:p119,Titiwangsa
+ocd-division/country:my/territory:14/ed:p120,Bukit Bintang
+ocd-division/country:my/territory:14/ed:p121,Lembah Pantai
+ocd-division/country:my/territory:14/ed:p122,Seputeh
+ocd-division/country:my/territory:14/ed:p123,Cheras
+ocd-division/country:my/territory:14/ed:p124,Bandar Tun Razak
+ocd-division/country:my/territory:16/ed:p125,Putrajaya
 ocd-division/country:my/state:05/ed:p126,Jelebu
 ocd-division/country:my/state:05/ed:p127,Jempol
 ocd-division/country:my/state:05/ed:p128,Seremban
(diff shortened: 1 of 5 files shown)


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-my/federal_states_terriroties.csv
```diff
@@ -0,0 +1,17 @@
...
+ocd-division/country:my/state:05,Negeri Sembilan
+ocd-division/country:my/state:06,Pahang
+ocd-division/country:my/state:07,Pulau Pinang
+ocd-division/country:my/state:08,Perak
+ocd-division/country:my/state:09,Perlis
+ocd-division/country:my/state:12,Sabah
+ocd-division/country:my/state:13,Sarawak
+ocd-division/country:my/state:10,Selangor
+ocd-division/country:my/state:11,Terengganu
+ocd-division/country:my/federal_territory:14,Wilayah Persekutuan Kuala Lumpur
```
jloutsenhizer, line 15: What do you think about using `territory` instead of `federal_territory` for some consistency with other country OCD IDs?

identifiers/country-my/federal_states_terriroties.csv
```diff
@@ -0,0 +1,17 @@
+id,name
```
jloutsenhizer, line 1: Please sort this file by OCD ID

identifiers/country-my/constituencies.csv
```diff
@@ -0,0 +1,223 @@
...
+ocd-division/country:my/federal_territory:16/ed:p125,Putrajaya
+ocd-division/country:my/state:05/ed:p126,Jelebu
+ocd-division/country:my/state:05/ed:p127,Jempol
+ocd-division/country:my/state:05/ed:p128,Seremban
+ocd-division/country:my/state:05/ed:p129,Kuala Pilah
+ocd-division/country:my/state:05/ed:p130,Rasah
+ocd-division/country:my/state:05/ed:p131,Rembau
+ocd-division/country:my/state:05/ed:p132,Telok Kemang
+ocd-division/country:my/state:05/ed:p133,Tampin
+ocd-division/country:my/state:04/ed:p134,Masjid Tanah
```
jloutsenhizer, line 135: I believe p134 - p139 are all within Malacca

identifiers/country-my/constituencies.csv
```diff
@@ -0,0 +1,223 @@
+id,name
```
jloutsenhizer, line 1: I'm a bit torn on whether to leave the nesting as is or to change it.

On one hand, it sounds like the Election Commission of Malaysia is in charge of all redistricting: https://en.wikipedia.org/wiki/Election_Commission_of_Malaysia#Functions

However, it also appears that, by law, each state/territory has a set number of constituencies so they will never cross those boundaries: https://web.archive.org/web/20090726120429/http://www.spr.gov.my/index_files/proses_praya/parlimen_dun.asp

So I'm leaning toward keeping the nesting of the electoral districts under each state, but would be intersted what others think.

jloutsenhizer (changes requested):

identifiers/country-my/federal_states_terriroties.csv
novitam: SGTM. Modified.

identifiers/country-my/federal_states_terriroties.csv
novitam: Done

identifiers/country-my/constituencies.csv
novitam: You are right, thanks for noticing this!

identifiers/country-my/constituencies.csv
novitam: Maybe @jpmckinney for opinions on the nesting?

jpmckinney: I think elsewhere we always have national/federal electoral districts hang off the country. In this case, they are also numbered to be unique at that level. If it were like the US where they are numbered within each state, there might be a reason to keep the state level.

jpmckinney: (This is assuming that the numbers for states and districts are from an official source.)

novitam: I did check official sources and noticed that they always use the numbers alongside the name, so I decided to keep the numbers as the unique identifiers. See:

Full list: https://www.spr.gov.my/sites/default/files/Senarai%20BPR%2010.09.2019.pdf

Other documents from the official site that list always the number + name:
- [notice of candidate name & official drawing for election #14](https://www.spr.gov.my/sites/default/files/MAKLUMAT%20PUSAT%20PENAMAAN%20CALON%20%28PPC%29%20%20PUSAT%20PENJUMLAHAN%20RASMI%20UNDI%20%28PPRU%29%20PRU%20KE-14%20-%20UPDATE%2018.4.2018_0.pdf)
- [re-examination report for sabah](https://www.spr.gov.my/sites/default/files/Laporan%20Sabah%20FINAL.pdf)
- [other areas of the site](https://www.spr.gov.my/ms/infografik-pilihan-raya)

jpmckinney (approved):
This looks structurally correct. I'm on vacation next week, so I'll approve now.

novitam:
@jloutsenhizer please re-review

jloutsenhizer (approved):
Changes look good to me

PR #275, opened 2021-09-29
Update census gazetteer file in country-us README

crystaljg:
The old link was broken so I searched the census website and replaced it with what I believe is the correct link. However, I'm not familiar with what the old link looked like, so hopefully this is the correct one!


jpmckinney:
Thanks!

PR #292, opened 2022-02-07
Update AU electorates for House of Representatives.

novitam:
Updated list of AU electoral divisions: https://en.wikipedia.org/wiki/Divisions_of_the_Australian_House_of_Representatives

Note: I am not changing the types here, and maintaining the use of `federal_electorate` for backward compatibilities.

### DIFF
--- a/identifiers/country-au/federal_electorates.csv
+++ b/identifiers/country-au/federal_electorates.csv
@@ -1,151 +1,162 @@
-id,name
-ocd-division/country:au/state:sa/federal_electorate:adelaide,South Australia - Federal House of Representatives Adelaide electorate
-ocd-division/country:au/state:vic/federal_electorate:aston,Victoria - Federal House of Representatives Aston electorate
-ocd-division/country:au/state:vic/federal_electorate:ballarat,Victoria - Federal House of Representatives Ballarat electorate
-ocd-division/country:au/state:nsw/federal_electorate:banks,New South Wales - Federal House of Representatives Banks electorate
-ocd-division/country:au/state:sa/federal_electorate:barker,South Australia - Federal House of Representatives Barker electorate
-ocd-division/country:au/state:nsw/federal_electorate:barton,New South Wales - Federal House of Representatives Barton electorate
-ocd-division/country:au/state:tas/federal_electorate:bass,Tasmania - Federal House of Representatives Bass electorate
-ocd-division/country:au/state:vic/federal_electorate:batman,Victoria - Federal House of Representatives Batman electorate
-ocd-division/country:au/state:vic/federal_electorate:bendigo,Victoria - Federal House of Representatives Bendigo electorate
-ocd-division/country:au/state:nsw/federal_electorate:bennelong,New South Wales - Federal House of Representatives Bennelong electorate
-ocd-division/country:au/state:nsw/federal_electorate:berowra,New South Wales - Federal House of Representatives Berowra electorate
-ocd-division/country:au/state:qld/federal_electorate:blair,Queensland - Federal House of Representatives Blair electorate
-ocd-division/country:au/state:nsw/federal_electorate:blaxland,New South Wales - Federal House of Representatives Blaxland electorate
-ocd-division/country:au/state:qld/federal_electorate:bonner,Queensland - Federal House of Representatives Bonner electorate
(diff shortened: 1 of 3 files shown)


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-au/federal_electorates.csv
```diff
@@ -1,151 +1,162 @@
...
+ocd-division/country:au/state:sa/federal_electorate:barker,South Australia - Federal House of Representatives Barker electorate,,
+ocd-division/country:au/state:sa/federal_electorate:boothby,South Australia - Federal House of Representatives Boothby electorate,,
+ocd-division/country:au/state:sa/federal_electorate:grey,South Australia - Federal House of Representatives Grey electorate,,
+ocd-division/country:au/state:sa/federal_electorate:hindmarsh,South Australia - Federal House of Representatives Hindmarsh electorate,,
+ocd-division/country:au/state:sa/federal_electorate:kingston,South Australia - Federal House of Representatives Kingston electorate,,
+ocd-division/country:au/state:sa/federal_electorate:makin,South Australia - Federal House of Representatives Makin electorate,,
+ocd-division/country:au/state:sa/federal_electorate:mayo,South Australia - Federal House of Representatives Mayo electorate,,
+ocd-division/country:au/state:sa/federal_electorate:port_adelaide,South Australia - Federal House of Representatives Port Adelaide electorate,,2019
```
jloutsenhizer, line 89: Are there more specific dates we can use than just the year?

identifiers/country-au/federal_electorates.csv
```diff
@@ -1,151 +1,162 @@
...
+ocd-division/country:au/state:nsw/federal_electorate:hunter,New South Wales - Federal House of Representatives Hunter electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:kingsford_smith,New South Wales - Federal House of Representatives Kingsford Smith electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:lindsay,New South Wales - Federal House of Representatives Lindsay electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:lyne,New South Wales - Federal House of Representatives Lyne electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:macarthur,New South Wales - Federal House of Representatives Macarthur electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:mackellar,New South Wales - Federal House of Representatives Mackellar electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:macquarie,New South Wales - Federal House of Representatives Macquarie electorate,,
+ocd-division/country:au/state:nsw/federal_electorate:mcmahon,New South Wales - Federal House of Representatives McMahon electorate,2010,
```
jloutsenhizer, line 30: I see some validFrom dates added for 2009 and 2010 electorates. Is there a specific reason to do this? I think it'd be reasonable to either have the validFrom dates added for districts added after this file was first created, or for all districts. The latter would be a great deal of work with little benefit, so  my preference would be for the former.

jloutsenhizer (changes requested):
Main issue is the specificity of the dates used in `validFrom` and `validTo`. I checked against https://en.wikipedia.org/wiki/Divisions_of_the_Australian_House_of_Representatives to check the number of added and removed districts for each year matched up.

identifiers/country-au/federal_electorates.csv
novitam: Thanks for calling it out. When cross checking with https://www.aec.gov.au/Electorates/Redistributions/dates.htm, the dates don't line up with the WIkipedia list, so I modified it to match the dates that are available in the AEC website.

The exact dates are available on the site by selecting the state -> Media Releases > "Names and boundaries of federal electoral divisions in <state> decided". For example: https://www.aec.gov.au/media/2021/06-28.htm

identifiers/country-au/federal_electorates.csv
novitam: Done

jloutsenhizer (approved):

novitam:
@jpmckinney please review, thanks!

jpmckinney (approved):
@jloutsenhizer already did the checks I was going to do (total numbers and validity dates), so approving :)

novitam:
Thanks both! :) @jloutsenhizer or @jpmckinney can you please merge this pull request?

PR #295, opened 2022-03-07
Add OCD ID's for Kenya

robinkiplangat:
Adding lists for the following administrative divisions of Kenya:

* counties: contains Contains 47 counties of Kenya
* national_assembly: Contains 290 constituencies in Kenya for Kenyan National Assembly
* county_assembly: Contains 1450 wards in Kenya for the 47 County Assemblies.

### DIFF
--- a/identifiers/country-ke/README.MD
+++ b/identifiers/country-ke/README.MD
@@ -1,0 +1,14 @@
+OCD Division Identifier
+---
+
+This folder `country-ke` lists the following administrative divisions of Kenya:
+
+* counties: contains Contains 47 counties of Kenya
+* national_assembly: Contains 290 constituencies in Kenya for Kenyan National Assembly
+* county_assembly: Contains 1450 wards in Kenya for the 47 County Assemblies. 
+
+
+Counties represented  based on the ISO 3166-2 Used sources [ISO 3166-2:KE](https://en.wikipedia.org/wiki/ISO_3166-2:KE)
+
+
+Constituencies & Wards  are represented as _ed_, and _ward_  respectively , sourced from [IEBC](https://www.iebc.or.ke/uploads/resources/AHxYKqSUn0.pdf)
--- a/identifiers/country-ke/counties.csv
+++ b/identifiers/country-ke/counties.csv
@@ -1,0 +1,48 @@
+id,name
+ocd-division/country:ke/county:ke-01,Mombasa
+ocd-division/country:ke/county:ke-02,Kwale
+ocd-division/country:ke/county:ke-03,Kilifi
+ocd-division/country:ke/county:ke-04,Tana River
+ocd-division/country:ke/county:ke-05,Lamu
+ocd-division/country:ke/county:ke-06,Taita Taveta
+ocd-division/country:ke/county:ke-07,Garissa
+ocd-division/country:ke/county:ke-08,Wajir
+ocd-division/country:ke/county:ke-09,Mandera
+ocd-division/country:ke/county:ke-10,Marsabit
+ocd-division/country:ke/county:ke-11,Isiolo
+ocd-division/country:ke/county:ke-12,Meru
+ocd-division/country:ke/county:ke-13,Tharaka - Nithi
+ocd-division/country:ke/county:ke-14,Embu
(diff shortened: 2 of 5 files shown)


identifiers/country-ke/counties.csv
```diff
@@ -0,0 +1,48 @@
+id,name
+ocd-division/country:ke/county:ke-01,Mombasa
```
jloutsenhizer, line 2: We should drop the `ke-` prefix. It's not included in other countries where we use ISO 3166-2 codes

sguenther85: +1 for drop ke-

identifiers/country-ke/README.MD
```diff
@@ -0,0 +1,14 @@
...
+---
+
+This folder `country-ke` lists the following administrative divisions of Kenya:
+
+* counties: contains Contains 47 counties of Kenya
+* national_assembly: Contains 290 constituencies in Kenya for Kenyan National Assembly
+* county_assembly: Contains 1450 wards in Kenya for the 47 County Assemblies. 
+
+
+Counties represented  based on the ISO 3166-2 Used sources [ISO 3166-2:KE](https://en.wikipedia.org/wiki/ISO_3166-2:KE)
```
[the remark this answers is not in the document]
jloutsenhizer: Please udpate this link to point to the iso.org resource instead of wikipedia.

identifiers/country-ke/counties.csv
jloutsenhizer: it looks like we still have a `ke` prefix on the county identifiers. I'd recommend dropping those so that the identifiers are just the numbers from the ISO 3166-2 codes.

identifiers/country-ke/county_assembly.csv
```diff
@@ -0,0 +1,1451 @@
+id,name
+ocd-division/country:ke/county:ke01/ed:baringo_central/ward:ewalel_chapchap,Ewalel-Chapchap
```
jloutsenhizer, line 2: I haven't thoroughly researched this but I have a question about the wards:

Are they really subdivided from the national assembly districts? In other words, if there is a redistricting of the assembly districts in a county, is it valid to have wards move from one assembly district to another?

jloutsenhizer (changes requested):
I did a spot check for correctness:
- the county ISO codes match up
- the number of national assembly districts look correct
- the number of countyy assembly wards looks correct

I have a couple additional questions to address, but I think this is in pretty good shape.

identifiers/country-ke/county_assembly.csv
robinkiplangat: Within the current administrative structure, wards are subdivided from national assembly districts, and the delimitation of electoral units per constitution ought to be reviewed at intervals of roughly between  8  to 12 years depending on population quotas. 

On the uploaded data, some Wards bear similar names due to the fact that these are localized regions and on occasion, names tend to be replicated eg Biashara Ward appears in 3 Constituencies (Ruiru, Naivasha & Nakuru Town East)

jloutsenhizer: Thanks for the additional details, but I don't think this answers my question.

Here's a more concrete hypothetical:
Suppose a redistricting happens and the size of Baringo South  `ocd-division/country:ke/county:ke01/ed:baringo_south` increases. With this change in size The number of wards in Baringo South increases from 4 to 5 and the number of wards in Baringo Central decreases from 5 to 4.
Let's suppose Tenges, `ocd-division/country:ke/county:ke01/ed:baringo_central/ward:tenges` is the ward removed from Baringo Central and added to Baringo South.

In this situation should the Tenges Ward be considered the same Ward in the County Government, or would it be considered a new Ward since it has moved to a new electoral district?

azuser:
Is there any update expected on this PR ? Otherwise, I can patch it and continue working on it.

marisviana:
Is this PR still being reviewed? If not, I can patch and continue the work. 

Thanks!

jloutsenhizer:
@robinkiplangat Are you planning on continuing work on this PR?

PR #296, opened 2022-03-07
Nesting Gallatin, MO under county:daviess

afsmythe:
Gallatin, MO is the county seat of Daviess County and there are instances of addresses in Gladstone, MO being incorrectly assigned to the existing `state:mo/place:gallatin` OCDID. Gladstone is a part of Clay County, which contains a [now inactive](https://en.wikipedia.org/wiki/Gallatin_Township,_Clay_County,_Missouri) Gallatin Township. 

This PR is an attempt to disambiguate the active Gallatin City (in Daviess County) with the inactive Gallatin Township (in Clay County).


jloutsenhizer:
I don't think I'm understanding what this PR is trying to solve. Is the issue that there are addresses in Gladstone county that still use the inactive Gallatin township as the town in the address instead of Gladstone?

PR #299, opened 2022-04-04
Update Malaysia constituencies.

novitam:
There has been some updates in the federal constituencies of Malaysia, however, the updates were made to the name and they keep the IDs of the constituencies. Hence, I didn't add the ValidTo, ValidFrom columns since the ids remain the same. https://www.freemalaysiatoday.com/category/nation/2018/03/28/ec-proposes-renaming-12-parliamentary-24-state-seats/

For example:
- P132: Port Dickson constituency was created in 2018 to replace Telok Kemang. https://en.wikipedia.org/wiki/Port_Dickson_(federal_constituency)
- P188: Lahad Datu constituency was created in 2019 to replace Silam. https://en.wikipedia.org/wiki/Lahad_Datu_(federal_constituency)


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-my.csv
```diff
@@ -72,7 +72,7 @@ ocd-division/country:my/state:05/ed:p128,Seremban
 ocd-division/country:my/state:05/ed:p129,Kuala Pilah
 ocd-division/country:my/state:05/ed:p130,Rasah
 ocd-division/country:my/state:05/ed:p131,Rembau
-ocd-division/country:my/state:05/ed:p132,Telok Kemang
+ocd-division/country:my/state:05/ed:p132,Port Dickson
```
jloutsenhizer, line 75: I'm not sure if we should continue to use the district numbering to identify these districts. From what I can tell these aren't renaming of the existing districts but the districts actually being replaced. Example:
- https://ms.wikipedia.org/wiki/Telok_Kemang_(kawasan_persekutuan)
- https://ms.wikipedia.org/wiki/Port_Dickson_(kawasan_persekutuan)

So I don't think it would be valid to say that an officeholder that represented Telok Kemang one term and then Port Dickson the next is representing the same district across both terms, and with the current identifiers we can't mark Telok Kemang with a `validTo` date.

What do you think about using the name as the electoral district identiifer?

Looping in @jpmckinney if he has any thoughts on this as well.

jloutsenhizer (changes requested):

identifiers/country-my.csv
jpmckinney: For context, in the US, there are states (if I remember correctly!) with congressional districts 1, 2, 3, etc. They don't have names – only numbers – and a redistricting *could* change their boundary entirely – from one part of the state to another. To preserve the identity of CD 1 for a given period, we *could* suffix the number with the year of the redistricting. By not adding this suffix, we treat CD 1 as persisting over time. Since congressional districts (given the possibility of being redrawn on an entirely separate territory), have no real identity (whether at a given time or across time), this seems like an okay compromise.

In the case of Malaysia, as I understand, the districts have names and they can come into and go out of existence (unlike the plain numbered districts in the US). However, the data source (what is the data source?), I guess, assigns sequential numbers each time, with no concern for reassigning numbers. In that case, I think it's better to use the names for IDs.

novitam: Thanks both. I have modified the files and use the district name as the id, and adding aliases.csv for backwards compatability. PTAL, thanks!

jloutsenhizer (changes requested):
Thanks for updating to use the names in the identifiers A couple changes needed:
1. I don't think we need the `ed:p#` OCD IDs the aliases + the new ones should be sufficient
2. I noticed some of the consituencies that were removed in 2018 are present in the files. Please add `validTo` dates to them so that it's clear they are past constituencies.

identifiers/country-my/house_of_representatives.csv
```diff
@@ -0,0 +1,232 @@
...
+ocd-division/country:my/state:12/ed:libaran,Libaran,,
+ocd-division/country:my/state:12/ed:papar,Papar,,
+ocd-division/country:my/state:12/ed:penampang,Penampang,,
+ocd-division/country:my/state:12/ed:pensiangan,Pensiangan,,
+ocd-division/country:my/state:12/ed:putatan,Putatan,,
+ocd-division/country:my/state:12/ed:ranau,Ranau,,
+ocd-division/country:my/state:12/ed:sandakan,Sandakan,,
+ocd-division/country:my/state:12/ed:semporna,Semporna,,
+ocd-division/country:my/state:12/ed:sepanggar,Sepanggar,,
+ocd-division/country:my/state:12/ed:silam,Silam,,
```
jloutsenhizer, line 183: It looks like Silam was removed in 2019: https://en.wikipedia.org/wiki/Silam_(federal_constituency)

identifiers/country-my/house_of_representatives.csv
```diff
@@ -0,0 +1,232 @@
...
+ocd-division/country:my/state:13/ed:igan,Igan,,
+ocd-division/country:my/state:13/ed:julau,Julau,,
+ocd-division/country:my/state:13/ed:kanowit,Kanowit,,
+ocd-division/country:my/state:13/ed:kapit,Kapit,,
+ocd-division/country:my/state:13/ed:kota_samarahan,Kota Samarahan,,
+ocd-division/country:my/state:13/ed:lanang,Lanang,,
+ocd-division/country:my/state:13/ed:lawas,Lawas,,
+ocd-division/country:my/state:13/ed:limbang,Limbang,,
+ocd-division/country:my/state:13/ed:lubok_antu,Lubok Antu,,
+ocd-division/country:my/state:13/ed:mambong,Mambong,,
```
jloutsenhizer, line 204: It looks like Mambong was removd in 2016: https://en.wikipedia.org/wiki/Mambong_(federal_constituency)

jloutsenhizer (changes requested):
Two districts need end dates still, then lgtm

identifiers/country-my/house_of_representatives.csv
novitam: done

identifiers/country-my/house_of_representatives.csv
novitam: done

jloutsenhizer (approved):

azuser:
@jpmckinney can you please review / approve this PR ?

PR #302, opened 2022-04-05
Add Ladakh as a Union Territory

novitam:
Add Ladakh as a Union Territory, it should be separate from Jammu and Kashmir since 2019-10-31.

https://en.wikipedia.org/wiki/Jammu_and_Kashmir_(union_territory)
https://en.wikipedia.org/wiki/Ladakh


novitam:
@jloutsenhizer PTAL, thanks!

jloutsenhizer (approved):

novitam:
@jpmckinney please review, thanks!

PR #304, opened 2022-04-12
Update ocd-ids that were duplicated, and make them unique.

novitam:
There were unfortunately duplicated ocd-ids in https://github.com/opencivicdata/ocd-division-ids/pull/257 because during the review process, we removed the state/provice/cities and made each district hang off the country:kr instead. For some reason this was not caught until recently.

In this PR, I am updating the duplicated ids to a unique one, and also sorting the file by the id.

Thanks,
Novita

### DIFF
--- a/identifiers/country-kr.csv
+++ b/identifiers/country-kr.csv
@@ -17,15 +17,12 @@ ocd-division/country:kr/ed:bucheon_city_1,Bucheon City First
 ocd-division/country:kr/ed:bucheon_city_2,Bucheon City Second
 ocd-division/country:kr/ed:bucheon_city_3,Bucheon City Third
 ocd-division/country:kr/ed:bucheon_city_fourth,Bucheon City Fourth
-ocd-division/country:kr/ed:buk_district,Buk District
-ocd-division/country:kr/ed:buk_district_and_gangseo_district_1,Buk District and Gangseo District First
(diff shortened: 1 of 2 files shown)


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-kr.csv
```diff
@@ -18,10 +18,12 @@ ocd-division/country:kr/ed:bucheon_city_2,Bucheon City Second
 ocd-division/country:kr/ed:bucheon_city_3,Bucheon City Third
 ocd-division/country:kr/ed:bucheon_city_fourth,Bucheon City Fourth
 ocd-division/country:kr/ed:buk_district,Buk District
-ocd-division/country:kr/ed:buk_district_1,Buk District First
-ocd-division/country:kr/ed:buk_district_2,Buk District Second
 ocd-division/country:kr/ed:buk_district_and_gangseo_district_1,Buk District and Gangseo District First
 ocd-division/country:kr/ed:buk_district_and_gangseo_district_2,Buk District and Gangseo District Second
+ocd-division/country:kr/ed:bukd_district_1,Buk District First (Daegu)
```
jloutsenhizer, line 23: instead of attaching a letter to the end of the first word in the district name, what if we include the city/province name in the front of the district ID. So in this case it'd be: `ed:daegu_buk_district_1`

novitam: Done

jloutsenhizer (approved):

PR #307, opened 2022-05-13
Update US state legislative districts for 2020 census redistricting

MiniCodeMonkey:
This adds new districts (and flag some as obsolete).

The changes affect the following states: MD, MA, VT, WV, WY

We (Geocodio) has collected redistricted boundaries state-by-state and matched them to OCD-ids, this PR is part of this effort.

### DIFF
--- a/identifiers/country-us/census_autogenerated/us_sldl.csv
+++ b/identifiers/country-us/census_autogenerated/us_sldl.csv
@@ -4872,6 +4872,8 @@ ocd-division/country:us/state:wy/sldl:57,Wyoming State House district 57
 ocd-division/country:us/state:wy/sldl:58,Wyoming State House district 58
 ocd-division/country:us/state:wy/sldl:59,Wyoming State House district 59
 ocd-division/country:us/state:wy/sldl:60,Wyoming State House district 60
+ocd-division/country:us/state:wy/sldl:61,Wyoming State House district 61
+ocd-division/country:us/state:wy/sldl:62,Wyoming State House district 62
 ocd-division/country:us/territory:pr/sldl:1,Puerto Rico State House district 1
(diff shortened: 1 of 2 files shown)


MiniCodeMonkey:
This brings all states except for the following up to date:

MO House
MO Senate
MT House
MT Senate
NH House
NH Senate
SC Senate

jdmgoogle:
Thanks for kicking off these updates.

Instead of adding "(obsolete)" to the names of districts which are going away, please add and populate the `validTo` field. E.g., as in [New Zealand](https://github.com/opencivicdata/ocd-division-ids/blob/master/identifiers/country-nz.csv).

MiniCodeMonkey:
@jdmgoogle That would be ideal! But I simply don't know the dates state-by-state.

jdmgoogle:
For the purposes here I think the date is whenever the new terms start in that legislative body. [This article](https://ballotpedia.org/When_state_legislators_assume_office_after_a_general_election) on Ballotpedia is the best resource that I've found ("When state legislators assume office after a general election")

nuttingd1:
For the Chittenden VT senate districts, can we get these type-id’s hyphenated instead of underscored like `sldu:chittenden-central` instead of `sldu:chittenden_central`? Hyphens seem to reflect either appending one common language name to another, or adding a qualifying value to a common language district description in VT. Looking at the old senate ocd-ids and this [doc](https://outside.vermont.gov/dept/sos/Elections%20Division/election%20info%20and%20resources/parties/house%20and%20senate%20districts/SENATE%20REDISTRICTING%202022.pdf) describing the new senate districts.

MiniCodeMonkey:
Continuing here: https://github.com/opencivicdata/ocd-division-ids/pull/330

PR #311, opened 2022-06-13
Update us_congressional_districts.csv

ballotpedia:
Adding at-large congressional positions from U.S. Territories

### DIFF
--- a/identifiers/country-us/census_autogenerated/us_congressional_districts.csv
+++ b/identifiers/country-us/census_autogenerated/us_congressional_districts.csv
@@ -435,6 +435,11 @@ ocd-division/country:us/state:wi/cd:8,Wisconsin's 8th congressional district,cd-
 ocd-division/country:us/state:wv/cd:1,West Virginia's 1st congressional district,cd-5401,
 ocd-division/country:us/state:wv/cd:2,West Virginia's 2nd congressional district,cd-5402,
 ocd-division/country:us/state:wy/cd:at-large,Wyoming's at-large congressional district,,
+ocd-division/country:us/territory:as/cd:at-large,American Samoa's at-large congressional district,cd-6098,
+ocd-division/country:us/territory:gu/cd:at-large,Guam's at-large congressional district,cd-6698,
+ocd-division/country:us/territory:mp/cd:at-large,Northern Mariana Islands's at-large congressional district,cd-6998,
+ocd-division/country:us/territory:pr/cd:at-large,Puerto Rico's at-large congressional district,cd-7298,
+ocd-division/country:us/territory:vi/cd:at-large,Virgin Islands's at-large congressional district,cd-7898,
 ocd-division/country:us/state:il/cd:19,Illinois's 19th congressional district (obsolete as of 2012),cd-1719,2013-01-03
 ocd-division/country:us/state:ia/cd:5,Iowa's 5th congressional district (obsolete as of 2012),cd-1905,2013-01-03
 ocd-division/country:us/state:la/cd:7,Louisiana's 7th congressional district (obsolete as of 2012),cd-2207,2013-01-03
--- a/identifiers/country-us/census_autogenerated/us_congressional_districts.csv
+++ b/identifiers/country-us/census_autogenerated/us_congressional_districts.csv
@@ -437,9 +437,9 @@ ocd-division/country:us/state:wv/cd:2,West Virginia's 2nd congressional district
 ocd-division/country:us/state:wy/cd:at-large,Wyoming's at-large congressional district,,
 ocd-division/country:us/territory:as/cd:at-large,American Samoa's at-large congressional district,cd-6098,
 ocd-division/country:us/territory:gu/cd:at-large,Guam's at-large congressional district,cd-6698,
-ocd-division/country:us/territory:mp/cd:at-large,Northern Mariana Islands's at-large congressional district,cd-6998,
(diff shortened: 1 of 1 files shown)


rsimoes:
May we additionally get numeric OCDIDs for these (`/cd:1`), for the sake of uniformity and also durability (given, e.g., `state:mt/cd:at-large` is no longer usable for current data)?

jloutsenhizer (changes requested):
Overall this looks good to me. Could you run `compile.py` to update the `country-us.csv` file?

jloutsenhizer:
re @rsimoes - My understanding is that, at least in the US, at-large congressional districts are considered distinct from first congressional districts, which is why Montana's at-large congressional district will no longer be a valid OCD ID after this year's elections. If in the future Montana goes back to a single congressional district then the first and second district OCD IDs would no longer be valid and the at-large congressional district OCD ID would once again be valid.

rsimoes:
> re @rsimoes - My understanding is that, at least in the US, at-large congressional districts are considered distinct from first congressional districts, which is why Montana's at-large congressional district will no longer be a valid OCD ID after this year's elections. If in the future Montana goes back to a single congressional district then the first and second district OCD IDs would no longer be valid and the at-large congressional district OCD ID would once again be valid.

There's a convention in the repository--inconsistently followed--to have a pair of aliased districts, one with "cd:at-large" and a second with "cd:1". The `sameAs` column isn't utilized for these but likely should be. From a year ago:

image

Personally I'd like this to be continued, but at the very least we should decide whether the above "cd:1" districts should be kept, and if not, removed or marked in some way as obsolete.

jloutsenhizer:
Sorry for the delayed response.

Do you have examples of using `cd:1` and `cd:at-large` aliased? We should probably be careful about that since in different systems the `at-large` and `first` districts are considerd distinct (like with congressional districts). The ones you show in your screenshot listed as historical districts: https://github.com/opencivicdata/ocd-division-ids/blob/98fd551c630ab0925b61c55595996eac9edb7072/identifiers/country-us/historical/unitedstates_legislators_autogenerated/historical_congressional_districts.csv

jloutsenhizer:
Looks like this might have fallen off of people's radars.

@ballotpedia could you run the compile.py for this PR?

ballotpedia:
@jloutsenhizer that is done!

jloutsenhizer (approved):

jloutsenhizer:
@rsimoes Do you still have concerns with this PR?

jloutsenhizer:
@ballotpedia it looks like this branch has conflicts. Could you follow the instructions to resolve them?

I think we can move forward with this PR at this point. @jpmckinney could you take a look for a final review?

identifiers/country-us/census_autogenerated/us_congressional_districts.csv
```diff
@@ -435,6 +435,11 @@ ocd-division/country:us/state:wi/cd:8,Wisconsin's 8th congressional district,cd-
 ocd-division/country:us/state:wv/cd:1,West Virginia's 1st congressional district,cd-5401,
 ocd-division/country:us/state:wv/cd:2,West Virginia's 2nd congressional district,cd-5402,
 ocd-division/country:us/state:wy/cd:at-large,Wyoming's at-large congressional district,,
+ocd-division/country:us/territory:as/cd:at-large,American Samoa's at-large congressional district,cd-6098,
+ocd-division/country:us/territory:gu/cd:at-large,Guam's at-large congressional district,cd-6698,
+ocd-division/country:us/territory:mp/cd:at-large,Northern Mariana Islands's at-large congressional district,cd-6998,
```
jpmckinney, line 440:
```suggestion
ocd-division/country:us/territory:mp/cd:at-large,Northern Mariana Islands' at-large congressional district,cd-6998,

```

identifiers/country-us/census_autogenerated/us_congressional_districts.csv
```diff
@@ -435,6 +435,11 @@ ocd-division/country:us/state:wi/cd:8,Wisconsin's 8th congressional district,cd-
 ocd-division/country:us/state:wv/cd:1,West Virginia's 1st congressional district,cd-5401,
 ocd-division/country:us/state:wv/cd:2,West Virginia's 2nd congressional district,cd-5402,
 ocd-division/country:us/state:wy/cd:at-large,Wyoming's at-large congressional district,,
+ocd-division/country:us/territory:as/cd:at-large,American Samoa's at-large congressional district,cd-6098,
+ocd-division/country:us/territory:gu/cd:at-large,Guam's at-large congressional district,cd-6698,
+ocd-division/country:us/territory:mp/cd:at-large,Northern Mariana Islands's at-large congressional district,cd-6998,
+ocd-division/country:us/territory:pr/cd:at-large,Puerto Rico's at-large congressional district,cd-7298,
+ocd-division/country:us/territory:vi/cd:at-large,Virgin Islands's at-large congressional district,cd-7898,
```
jpmckinney, line 442:
```suggestion
ocd-division/country:us/territory:vi/cd:at-large,Virgin Islands' at-large congressional district,cd-7898,

```

jpmckinney (commented):
Looks good minus possessive typos.

kcroo:
Hi @jpmckinney , I fixed those possessive typos and rebased to remove the merge conflicts. Please let me know if there is anything else!

jpmckinney (approved):

PR #314, opened 2022-06-13
Update us_state_courts.csv

ballotpedia:
Add Washington Court of Appeals subdivisions

### DIFF
--- a/identifiers/country-us/us_state_courts.csv
+++ b/identifiers/country-us/us_state_courts.csv
@@ -1549,6 +1549,15 @@ ocd-division/country:us/state:va/circuit_court:31,VA State Circuit Court - 31st 
 ocd-division/country:us/state:wa/court_of_appeals:1,WA State Court of Appeals - 1st Division
 ocd-division/country:us/state:wa/court_of_appeals:2,WA State Court of Appeals - 2nd Division
 ocd-division/country:us/state:wa/court_of_appeals:3,WA State Court of Appeals - 3rd Division
+ocd-division/country:us/state:wa/court_of_appeals:division_1_district_1,WA State Court of Appeals - 1st Division District 1
+ocd-division/country:us/state:wa/court_of_appeals:division_1_district_2,WA State Court of Appeals - 1st Division District 2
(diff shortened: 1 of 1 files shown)


jloutsenhizer (changes requested):
Looks good overall, please run `compile.py` to update the `country-us.csv`

ballotpedia:
Thanks for the comments @jloutsenhizer. Could we get some more guidance on the process for implementing these changes at some point? I am not totally clear on what sort of script you are asking us to run, or what kinds of approval is needed as we make these requests to the repo, so any summary of the steps involved would be helpful. Thanks! -Matt

jloutsenhizer:
Sorry for delays in replying. 

The script I'm referring to is: https://github.com/opencivicdata/ocd-division-ids/blob/master/scripts/compile.py

It takes the files within in country directory and updates the `country-xx.csv` file. All PRs should use that script to ensure those files are up to date. There are also automated tests which verify this has been run. [Example](https://github.com/opencivicdata/ocd-division-ids/runs/7047353769?check_suite_focus=true)

As for approvals, the typical process is that at least 2 approvals are required. Typically once someone has reviewed the PR and approved it, we'll pull in another reviewer that's eligible to approve and merge the PR. Let me know if there's more questions.

jloutsenhizer (approved):

jpmckinney:
Hmm, so we already have `ocd-division/country:us/state:wa/court_of_appeals:1` for WA State Court of Appeals - 1st Division, etc.

I think the districts should hang off these existing parents, so `ocd-division/country:us/state:wa/court_of_appeals:1/district:1`, etc.

jpmckinney:
If some software asks "How many court_of_appeals divisions are there in the State of Washington?", then after this PR, the answer would be 12. Right now, the answer is 3 (correct). I think the IDs need to be structured to answer this question correctly.

jloutsenhizer:
That's a fair point, I suppose either the types would have to be different on the district OCD IDs or nesting would also accomplish this. I think nesting looks cleaner so I agree with your recommendation

ballotpedia:
We've submitted a new pull request that addresses the requested changes https://github.com/opencivicdata/ocd-division-ids/pull/321

PR #318, opened 2022-09-07
Add assembly constituencies for Jammu & Kashmir union territory.

novitam:
Source: https://en.wikipedia.org/wiki/List_of_constituencies_of_the_Jammu_and_Kashmir_Legislative_Assembly


novitam:
@jloutsenhizer PTAL, thanks!

identifiers/country-in/districts.csv
```diff
@@ -692,6 +692,26 @@ ocd-division/country:in/territory:dl/district:se,South East Delhi
...
 ocd-division/country:in/territory:dl/district:wd,West Delhi
+ocd-division/country:in/territory:jk/district:an,Anantnag
+ocd-division/country:in/territory:jk/district:bn,Bandipora
+ocd-division/country:in/territory:jk/district:br,Baramulla
+ocd-division/country:in/territory:jk/district:bu,Budgam
+ocd-division/country:in/territory:jk/district:do,Doda
+ocd-division/country:in/territory:jk/district:ga,Ganderbal
+ocd-division/country:in/territory:jk/district:ja,Jammu
+ocd-division/country:in/territory:jk/district:ka,Kathua
+ocd-division/country:in/territory:jk/district:ki,Kishtawar
```
jloutsenhizer, line 703: Should this be spelled `Kishtwar`? https://en.wikipedia.org/wiki/Kishtwar_district

jloutsenhizer (changes requested):
Just one comment on a potential spelling issue. Otherwise this looks good to me

identifiers/country-in/districts.csv
novitam: You are right! Corrected.

novitam:
@jpmckinney can you please review? thanks!

jpmckinney (approved):
The change in sort order makes this very difficult to review. I'll just assume it's correct. In figure, try to separate sorting into a separate commit, so that I can ignore sorting-only changes.

Issue #323, opened 2022-11-01
Using district names in Canadian Federal Electoral District OCD-IDs


evannjw:
A new set of OCD-IDs are created every time there is redistricting creating a problem for how we represent districts contiguously across time. Rather than use the numeric federal district ids, which only represent a district’s identity for 10 years, we should use district names as they are more closely represent a district’s identity. Some concerns regarding the switch to using names as identifiers are that they can change between redistricting cycles (see https://www12.statcan.gc.ca/census-recensement/2011/ref/FED_name_changes-changements_noms_CEF-eng.cfm and https://openparliament.ca/bills/42-1/C-402/). When names are changed, new ids and aliases can be created so that identity is maintained through the canonical OCD-ID. When redistricting occurs, districts created after redistricting that don’t map to an old one (not the same name or renaming of a district) will be considered new districts and ids will be created for them. Old districts which no longer exist will have a validThrough date to indicate it is no longer active.

jpmckinney:
To understand the proposal:

- When X is renamed to Y, which is the canonical ID and which is the alias?
- If X is canonical: how far back do we go to establish the canonical ID? (1867? Easy enough with the source below.)
- How are collisions resolved across time (same name for different districts at different times)? For example, Ahuntsic from 1968 to 1979 vs 1988 to 2015. There are 213 names that repeat, up to 6 times (Victoria).

<details>

- Ahuntsic: 2
- Algoma: 2
- Argenteuil: 3
- Argenteuil--Deux-Montagnes: 2
- Argenteuil--Papineau--Mirabel: 2
- Bas-Richelieu--Nicolet--Bécancour: 2
- Battle River: 3
- Beauharnois: 3
- Beauharnois--Salaberry: 3
- Beauséjour: 2
- Berthier: 2
- Berthier--Maskinongé: 3
- Bourassa: 2
- Bow River: 3
- Bramalea--Gore--Malton: 2
- Brampton Centre: 2
- Brant: 2
- Brantford: 2
- Brome--Missisquoi: 3
- Bruce South: 2
- Bruce--Grey--Owen Sound: 2
- Burin--St. George's: 2
- Burrard: 2
- Calgary: 2
- Calgary Centre: 2
- Calgary East: 3
- Calgary West: 2
- Cape Breton South: 2
- Cape Breton--East Richmond: 2
- Cape Breton--The Sydneys: 2
- Cariboo: 2
- Carleton: 3
- Carleton--Gloucester: 2
[...]

PR #324, opened 2022-11-04
ca: Create new files for current, abolished and renamed OCD-IDs

jpmckinney:
This PR implements a naming policy co-created with @evannjw in #323.

In brief:

* All current divisions are given an ID, based on their name.
* All divisions that are abolished according to the Library of Parliament are given an ID, based on their name.
* If multiple current/abolished divisions have the same ID, later divisions are suffixed with the validFrom year.
* All divisions that were renamed according to the Library of Parliament are aliases (sameAs) to another ID.
* A province/territory:id pair is prefixed to the ed:id pair, since two provinces can contain the same district name.
* The most recent ID is the "canonical" ID. Previous renames are aliased to it. This ensures that renamings to respect language, culture, etc. (e.g. Three Rivers -> Trois Rivières) are reflected in the dataset, without giving primacy to the old name that might be colonial, etc.

To complete the PR:

* [ ] Add missing sameAs aliases (about 60 left to resolve either via code or manually)
* [ ] Do the TODOs in the code (also remove validFrom,validThrough from aliases CSV)
* [ ] Add sameAs aliases between the existing OCD-IDs and these new ones


jpmckinney:
Sounds good. Yes, I wanted to get agreement on what I had so far before completing the TODOs.

We can rename the old files (those IDs will be aliases to the new ones), and document the naming policy in the readme.

The Ruby scripts for those IDs would be removed, as the Python script is now responsible for federal IDs.

jpmckinney:
I can work on the TODOs in the Python script, if you can help with aliasing the current federal IDs (`ca_federal_electoral_districts-2013.csv` should be easy, `ca_federal_electoral_districts.csv` probably requires a lookup that accounts for name, validFrom, validThrough).

evannjw:
Sounds good, I'll also update the README and rename/clean up old files.

evannjw:
I don't seem to have access to push to this branch/repo. Here is a file containing aliases of current federal IDs. 
[aliases.csv](https://github.com/opencivicdata/ocd-division-ids/files/9964507/aliases.csv)

jpmckinney:
@evannjw I don't have much time this week for this PR, so if you can help with the other items, please have a look.

evannjw:
Sure I can wrap up the remaining TODOs here

evannjw:
Hi sorry, haven't actually had time to work on this, should be able to get to this tomorrow/early next week

jpmckinney:
Replaced by #327

PR #331, opened 2023-01-02
QC Redistribution 2022

DrMeers:
Please confirm it is correct/OK to remove the 2017 districts/scraper, and only maintain one set of districts going forward? I assumed we only maintain parallel sets when the redistribution is future-dated, but this seems to have been in effect since October 2022?


jpmckinney:
I'm not sure what this PR is for. There has been no redistricting since 2017, and there are already IDs for the 2017 redistricting. Closing pending clarification.

DrMeers:
> I'm not sure what this PR is for.

@jpmckinney the QC scraper is currently failing with a missing district: `pupa.exceptions.UnresolvedIdError: cannot resolve pseudo id to Post: ~{"label": "Camille-Laurin", "organization__classification": "legislature", "role": "MNA"}`

As per this PR, https://www.electionsquebec.qc.ca/ is now listing a different set of districts to the `-2017` ones, including "Camille-Laurin"

jpmckinney:
They probably just renamed a few districts. Make a PR preserving the suffixes.

jpmckinney:
@DrMeers I think you misunderstood. The fix was very simple. You just had to replace one URL, as shown here: https://github.com/opencivicdata/ocd-division-ids/commit/ccbcaf64b8bd1979891b5998833616646868f86a

Like I said, the province did not do a redistricting. An MNA simply passed a bill to rename a single district. That means we just change one row, not hundreds of rows...

DrMeers:
ah thanks @jpmckinney

PR #335, opened 2023-03-07
Created a file with the constituencies in Finland, and a README file …

marisviana:
…with more information and context around these changes


marisviana:
@jloutsenhizer Hey Justin, could you please take a look at the proposed changes for Finland? Thanks!

sguenther85:
Hi Mari,

ocd-division/country:fi/vaalipiirit:helsinki,Helsinki

we should use/prefer "ed" for electoral districts instead of "vaalipiirit". 
Source: https://en.wikipedia.org/wiki/Constituencies_of_Finland

The rest looks good!

marisviana:
Thanks Steve, just updated the file.

sguenther85:
Maybe rename the vaalipiirit.csv now to electoral_districts.csv like other ocd country files and then also sort the ocd paths alphabetically.

After that: @jloutsenhizer @jpmckinney do you two have anything else?
would be great if we could use the file for the election

marisviana:
Sounds good, thanks! I just applied these changes and I think now we are good to go. Just waiting for the reviewers to take a look, and as Steve mentioned, the idea is to use these for the upcoming elections.

identifiers/country-fi/README.md
```diff
@@ -0,0 +1,5 @@
+This folder lists the following [constituencies of Finaland](https://en.wikipedia.org/wiki/Constituencies_of_Finland):
```
jloutsenhizer, line 1: Could you update the READEME to cover the other files as well?

identifiers/country-fi/electoral_districts.csv
```diff
@@ -0,0 +1,14 @@
+id,name
+ocd-division/country:fi/ed:helsinki,Helsinki
+ocd-division/country:fi/ed:åland,Åland
+ocd-division/country:fi/ed:haeme,Häme
```
jloutsenhizer, line 4: Is there a reason to have the OCD ID use haeme instead of häme?

jloutsenhizer (changes requested):
Overall looks good, just a couple comments

marisviana:
Sounds good, thanks! I just updated the README file and applied your suggestions with regards to the OCD ID (makes more sense).

jloutsenhizer (approved):

jpmckinney (approved):
I approve.

Sidenote: I assume the numbers for the districts on Wikipedia are just to assist in reading the map, and are not an official number for each district (I don't see any evidence to that).

PR #337, opened 2023-03-22
Created a file for the departments in Uruguay, alongside a README fil…

marisviana:
…e that contains an explanation of these changes

### DIFF
--- a/identifiers/country-uy/departments.csv
+++ b/identifiers/country-uy/departments.csv
@@ -1,5 +1,4 @@
 id,name
-ocd-division/country:uy,Uruguay
 ocd-division/country:uy/department:uy-ar,Artigas
 ocd-division/country:uy/department:uy-ca,Canelones
 ocd-division/country:uy/department:uy-cl,Cerro Largo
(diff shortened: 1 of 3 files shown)


marisviana:
@jloutsenhizer Hey Justin, could you please take a look at the proposed changes for Uruguay? Thanks!

identifiers/country-uy/departments.csv
```diff
@@ -0,0 +1,20 @@
+id,name
+ocd-division/country:uy/department:uy-ar,Artigas
```
jloutsenhizer, line 2: I think we shuold drop the `uy-` prefex from the department identifiers since it's redundant with the `country:uy` portion.

marisviana: Sounds good, I just removed them.

jloutsenhizer (approved):

marisviana:
@jpmckinney, could you please take a look at these changes and see if they make sense? Thanks!

jpmckinney (approved):

PR #340, opened 2023-06-12
Expansion of the electoral list in spain

sguenther85:
For the upcoming election in Spain we need to extend the current list with a couple of subdivision which are necessary for a complete candidates and officeholder list of spain.

The former list is still correct, but for three provinces around the bealeares and canaries they use a couple of subdivision for the elections and their candidates. (parliament and senate)

Here is a list from the last election of 2019:
[2019 Candidaturas](http://www.juntaelectoralcentral.es/cs/jec/documentos/Generales_2019-R_Candidaturas%20proclamadas.pdf)

### DIFF
--- a/identifiers/country-es.csv
+++ b/identifiers/country-es.csv
@@ -37,7 +37,14 @@ ocd-division/country:es/autonomous_community:cm/province:gu,Guadalajara
 ocd-division/country:es/autonomous_community:cm/province:to,Toledo
 ocd-division/country:es/autonomous_community:cn,Canarias
 ocd-division/country:es/autonomous_community:cn/province:gc,Las Palmas
+ocd-division/country:es/autonomous_community:cn/province:gc-1,Gran Canaria
+ocd-division/country:es/autonomous_community:cn/province:gc-2,Lanzarote
+ocd-division/country:es/autonomous_community:cn/province:gc-3,Fuerteventura
 ocd-division/country:es/autonomous_community:cn/province:tf,Santa Cruz de Tenerife
+ocd-division/country:es/autonomous_community:cn/province:tf-1,Tenerife
+ocd-division/country:es/autonomous_community:cn/province:tf-2,La Gomera
+ocd-division/country:es/autonomous_community:cn/province:tf-3,El Hierro
+ocd-division/country:es/autonomous_community:cn/province:tf-4,La Palma
 ocd-division/country:es/autonomous_community:ct,Cataluña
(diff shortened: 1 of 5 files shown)


identifiers/country-es/README.md
jloutsenhizer, line 1: I think two spaces need to be used for indentend lists in markdown. Could you update this file accordingly?

identifiers/country-es/provincia.csv
```diff
@@ -26,7 +26,14 @@ ocd-division/country:es/autonomous_community:cm/province:cu,Cuenca
 ocd-division/country:es/autonomous_community:cm/province:gu,Guadalajara
 ocd-division/country:es/autonomous_community:cm/province:to,Toledo
 ocd-division/country:es/autonomous_community:cn/province:gc,Las Palmas
+ocd-division/country:es/autonomous_community:cn/province:gc-1,Gran Canaria
```
jloutsenhizer, line 29: My understanding is these additional OCD IDs aren't actually provinces, but instead districts used for elections right?

I think `ed` would be a more appropriate type to mark them as electoral districts and nest it at whatever level of government the decision of the creation of these electoral districts is being made. For example, if at the province-level they are deciding to subdivide into these electoral districts then the current nesting looks fine.

I also think it'd be better to have these in a separate file since they aren't provinces.

sguenther85:
Hi @jloutsenhizer. Could you have a second look? And if its all good a further second pserson?
Best Steve

jloutsenhizer:
@jpmckinney for a second review

identifiers/country-es.csv
```diff
@@ -53,6 +60,9 @@ ocd-division/country:es/autonomous_community:ga/province:or,Ourense
 ocd-division/country:es/autonomous_community:ga/province:po,Pontevedra
 ocd-division/country:es/autonomous_community:ib,Illes Balears
 ocd-division/country:es/autonomous_community:ib/province:pm,Baleares
+ocd-division/country:es/autonomous_community:ib/province:pm/ed:1,Mallorca
+ocd-division/country:es/autonomous_community:ib/province:pm/ed:2,Menorca
+ocd-division/country:es/autonomous_community:ib/province:pm/ed:3,Ibiza-Formentera
```
jpmckinney, line 65: I can't find this in the PDF. I see "Eivissa-Formentera"

sguenther85: "Eivissa" is the "Catalan" name for Ibiza (spanish). 
So i would keep Ibiza in this case.

jpmckinney (approved):

PR #341, opened 2023-06-15
Created a file for the counties and another one for the constituencie…

marisviana:
…s of Kenya, apart from a readme file with more context for these changes. 

There are 47 counties and 290 constituencies (represented as ed in the constituencies file).

### DIFF
--- a/identifiers/country-ke/counties.csv
+++ b/identifiers/country-ke/counties.csv
@@ -1,0 +1,48 @@
+id,name
+ocd-division/country:ke/county:01,Baringo
+ocd-division/country:ke/county:02,Bomet
+ocd-division/country:ke/county:03,Bungoma
+ocd-division/country:ke/county:04,Busia
+ocd-division/country:ke/county:05,Elgeyo/Marakwet
+ocd-division/country:ke/county:06,Embu
+ocd-division/country:ke/county:07,Garissa
+ocd-division/country:ke/county:08,Homa Bay
+ocd-division/country:ke/county:09,Isiolo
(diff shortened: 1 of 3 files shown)


marisviana:
@jloutsenhizer could you please take a look to see if these changes make sense? Thanks!

identifiers/country-ke/counties.csv
```diff
@@ -0,0 +1,48 @@
+id,name
+ocd-division/country:ke/county:01,Baringo
```
evannjw, line 2: I'd recommend using names instead of numbers in the identifier unless there's a good reason not to. So something like ocd-division/country:ke/county:baringo. Here and in counstituencies.csv. Otherwise looks good.

marisviana: I am using the ISO codes listed in here: https://www.iso.org/obp/ui/#iso:code:3166:KE 

So not sure if it would be ok to not apply them here (and just for context, I have been using ISO codes for other countries whenever they are available).

evannjw: Ah, sorry I missed that, for some reason I thought that was referencing the county name. Looks good to me then.

evannjw (approved):

jloutsenhizer (approved):
Checked the overall counts for the counties and constituency and spot checked a few. LGTM

jloutsenhizer:
@jpmckinney for an additional review

jpmckinney (approved):
Looks good. It's confusing that Wikipedia uses a different number than ISO for some counties.

PR #327, opened 2022-11-30
323 canada fed rename

evannjw:
Cleaned up conflicting aliases and completed TODOs on draft pr.
For missing sameAs, these were moved to abolished districts csv as there is no clear source to use to address these.

closes #323 (which followed from #320)
replaces #324

PR #347, opened 2023-08-02
Poland: add senate constituencies, updated congressional districts and README

zikowang:
For the upcoming election in Poland (expected on 31st October 2023) we need to add a file for the senate constituencies of Poland.

There were changes happened to some congressional districts in 2011 and 2019.

- the congressional districts were moved to an own file named `congressional_districts.csv`
- the electoral districts are now in the `electoral_districts.csv` file

A readme file for this folder was added as well.


zikowang:
@jloutsenhizer @jpmckinney could you have a look at the changes in poland, please?

evannjw (approved):

identifiers/country-pl/congressional_districts.csv
jloutsenhizer, line 1: It might be a little clearer if we name these files based on the corresponding body the districts correspond to. So perhaps `congressional_districts.csv` -> `sejm_constituencies.csv` and `electoral_districts.csv` -> `senate_districts.csv`

jpmckinney (commented):
It's looking good, with a few edits. I was initially confused because, effectively, electoral_districts.csv is renamed to congressional_district.csv, and then electoral_districts.csv is used for senate districts.

It would be clearer to rename `electoral_districts.csv` to `senate_districts.csv` – and while were doing that change `congressional_districts.csv` to `parliament_districts.csv` (Poland had a "Congress Poland" from 1815 to 1831, so it's confusing).

I don't think senate districts should be nested under parliament districts. They do have that correspondence, but there is no political hierarchy between these divisions (only a geographical hierarchy).

Also, Poland numbers Senate districts from 1-100, and names them "No.1 (Legnica)" etc. https://sejmsenat2019.pkw.gov.pl/sejmsenat2019/en/wyniki/senat/pl

For backwards compatibility, we'll keep the same choice of "cd" and identifier for parliamentary districts (though Poland also uses numbers here).

jloutsenhizer:
> Also, Poland numbers Senate districts from 1-100, and names them "No.1 (Legnica)" etc. https://sejmsenat2019.pkw.gov.pl/sejmsenat2019/en/wyniki/senat/pl

It does look like that is how they are primarily referred to based on poking around the 2019 election website and [2015 election website](https://parlament2015.pkw.gov.pl/351_Wyniki_Senat.html). The name that comes along with the number is the "seat of the district electoral commission" which presumably can change from election to election, as is the case with some of the ID updates in this PR.

It looks like these numbers can be reused over time. Example [Senate constituency no 25 (2001-2011)](https://pl.wikipedia.org/wiki/Okr%C4%99g_wyborczy_nr_25_do_Senatu_Rzeczypospolitej_Polskiej_(2001%E2%80%932011)) and [Senate consituency no 25 (2011-present)](https://pl.wikipedia.org/wiki/Okr%C4%99g_wyborczy_nr_25_do_Senatu_Rzeczypospolitej_Polskiej). 

So I'd propose using the senate elctoral district number + the year they were established. Which I believe would be 2011 for all of the current districts. I think this would be consistent with [the current proposal for Canadian districts](https://github.com/opencivicdata/ocd-division-ids/pull/327) which have a similar name re-use issue.

zikowang:
Thank you for your reviews.

Files will be renamed.

> So I'd propose using the senate electoral district number + the year they were established. Which I believe would be 2011 for all of the current districts. I think this would be consistent with [the current proposal for Canadian districts](https://github.com/opencivicdata/ocd-division-ids/pull/327) which have a similar name re-use issue.

Just to get it right, for example
```csv
ocd-division/country:pl/state:mz/cd:warszawa_i/ed:1,Warszawa I 1
```
should then be
```csv
ocd-division/country:pl/state:mz/ed:42-2011,Warszawa I 1
```
42 is the number of the senate district and 2011 is the year it was established.

jpmckinney:
Good catch on suffixing by year created. IDs like `ed:25-2011` look good.

> The name for all 6 senate districts in Warszawa are identical (Warszawa), so I would add a number. Warszawa EDs begins at number 40. So number 40 would be Warszawa 1, number 41 Warszawa 2 and so on.

Sources instead name districts as e.g. "No. 40 (Warszawa)", so let's do that instead of inventing new numbers.

States (provinces) in Poland have their own [elected legislatures](https://en.wikipedia.org/wiki/Voivodeship_sejmik), and so the `ed` should not hang off the `state` but off the `country` itself.

identifiers/country-pl/congressional_districts.csv
zikowang: Thank you for your review. The file are renamed.

zikowang:
> Sources instead name districts as e.g. "No. 40 (Warszawa)", so let's do that instead of inventing new numbers.
Good call. Done.

> States (provinces) in Poland have their own [elected legislatures](https://en.wikipedia.org/wiki/Voivodeship_sejmik), and so the `ed` should not hang off the `state` but off the `country` itself.

Eds now hanging off the `country` itself.

Can you review again, please.

jloutsenhizer (approved):

identifiers/country-pl/senate_districts.csv
```diff
@@ -0,0 +1,101 @@
...
+ocd-division/country:pl/ed:1-2011,No 1 (Legnica)
+ocd-division/country:pl/ed:2-2011,No 2 (Legnica)
+ocd-division/country:pl/ed:3-2011,No 3 (Legnica)
+ocd-division/country:pl/ed:4-2011,No 4 (Wałbrzych)
+ocd-division/country:pl/ed:5-2011,No 5 (Wałbrzych)
+ocd-division/country:pl/ed:6-2011,No 6 (Wrocław)
+ocd-division/country:pl/ed:7-2011,No 7 (Wrocław)
+ocd-division/country:pl/ed:8-2011,No 8 (Wrocław)
+ocd-division/country:pl/ed:9-2011,No 9 (Bydgoszcz)
+ocd-division/country:pl/ed:10-2011,No.10 (Bydgoszcz)
```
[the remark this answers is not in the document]
zikowang: Oh, how did i miss that! Fixed.

jpmckinney (approved):

PR #349, opened 2023-09-08
Fix Grand Isle, VT SLDU

jones-liam-j:
Fixes the [Vermont State Senate district "Grand Isle"](https://legislature.vermont.gov/Documents/SmallMaps/2022/Senate/Grand%20Isle.pdf) erroneously being marked as obsolete, and the non-existent "Grand Isle-Chittenden" (which is actually a [lower house district](https://legislature.vermont.gov/Documents/SmallMaps/2022/House/Grand%20Isle-Chittenden.pdf)) existing in its place.

According to [state reapportionment records](http://leg.state.vt.us/reapportionment/) this district has been named "Grand Isle" since at least 2002; Chittenden is a separate district (or three districts, as of 2022).


identifiers/country-us.csv
```diff
@@ -179171,8 +179171,7 @@ ocd-division/country:us/state:vt/sldu:chittenden-southeast,Vermont Chittenden So
 ocd-division/country:us/state:vt/sldu:essex,Vermont Essex State Senate district,,,,,,,,,,,2023-01-04
 ocd-division/country:us/state:vt/sldu:essex-orleans,Vermont Essex-Orleans State Senate district,,,2023-01-03,,sldu-50E-O,sldu-50E-O,vt-upper-essex-orleans,,,,
 ocd-division/country:us/state:vt/sldu:franklin,Vermont Franklin State Senate district,,,,,sldu-50FRA,sldu-50FRA,vt-upper-franklin,,,,
-ocd-division/country:us/state:vt/sldu:grand_isle,Vermont Grand Isle State Senate district (obsolete),,,,,sldu-50CGI,,,,,,
-ocd-division/country:us/state:vt/sldu:grand_isle-chittenden,Vermont Grand-Isle-Chittenden State Senate district,,,,,,sldu-50CGI,vt-upper-grand-isle-chittenden,,,,
```
jloutsenhizer, line 179175: Instead of completely deleting `ocd-division/country:us/state:vt/sldu:grand_isle-chittenden`, it should be added to `aliases.csv` to mark that it is the same-as `ocd-division/country:us/state:vt/sldu:grand_isle` since some systems might already be using it and relying on it.

jloutsenhizer (changes requested):
Overall I think this is a good change, once the erroneous ID is mapped to the correct ID in via a sameAs alias

PR #359, opened 2023-12-08
Add Districts MX

RodrigoRibeiroSantos:
Including missing districts from Mexico

### DIFF
--- a/identifiers/country-mx.csv
+++ b/identifiers/country-mx.csv
@@ -1,4 +1,4 @@
-id,name,validFrom,validTo
+id,name,validFrom,validThrough
 ocd-division/country:mx,Mexico
 ocd-division/country:mx/region:1,Primera
 ocd-division/country:mx/region:2,Segunda
--- a/identifiers/country-mx.csv
+++ b/identifiers/country-mx.csv
@@ -1,4 +1,4 @@
-id,name
+id,name,validFrom,validTo
 ocd-division/country:mx,Mexico
 ocd-division/country:mx/region:1,Primera
 ocd-division/country:mx/region:2,Segunda
@@ -18,7 +18,7 @@ ocd-division/country:mx/state:bc/fed:5,V Federal Electoral District of Baja Cali
 ocd-division/country:mx/state:bc/fed:6,VI Federal Electoral District of Baja California
 ocd-division/country:mx/state:bc/fed:7,VII Federal Electoral District of Baja California
 ocd-division/country:mx/state:bc/fed:8,VIII Federal Electoral District of Baja California
-ocd-division/country:mx/state:bc/fed:9,IX Federal Electoral District of Baja California
+ocd-division/country:mx/state:bc/fed:9,IX Federal Electoral District of Baja California,2024-06-02
 ocd-division/country:mx/state:bs,Baja California Sur
 ocd-division/country:mx/state:bs/fed:1,I Federal Electoral District of Baja California Sur
 ocd-division/country:mx/state:bs/fed:2,II Federal Electoral District of Baja California Sur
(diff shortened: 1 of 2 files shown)


evannjw (approved):

identifiers/country-mx/districts.csv
jloutsenhizer, line 1: My understanding is that There are 300 federal electoral districts in Mexico: https://pdba.georgetown.edu/ElecSys/Mexico/mexico.html

However, we have 307 after this change. Based on [this wiki page](https://en.wikipedia.org/wiki/Federal_electoral_districts_of_Mexico), my understanding is that redistricting last happened last year for the upcoming election next year. Could you update these with validFrom (for new districts) and validThrough (for removed districts) to reflect this?

jloutsenhizer (changes requested):
Overall looks good, I think we just need some `validFrom` and `validThrough` attributes to make the data a little clearer.

identifiers/country-mx/districts.csv
RodrigoRibeiroSantos: Added validFrom "2022-12-12" for new 7 districts.

jloutsenhizer: What's the significance of `2022-12-12` ? it looks like the next election is `2024-06-02`: https://en.wikipedia.org/wiki/2024_Mexican_general_election

I'd recommend using `2024-06-02` as the `validFrom` date, can you also add `validThrough` of `2024-06-02` for the 7 districts that are going away as well?

RodrigoRibeiroSantos: Changed the validFrom and validThrough to "2024-06-02". Thanks

jloutsenhizer: Sorry I think there's was a misunderstanding. The updates as is means each of these new districts is valid only on `2024-06-02`.

For `validThrough` I was requesting it be added to the 7 districts that are going away since we should have 300 valid districts on any given date.

jloutsenhizer:
@RodrigoRibeiroSantos are there any updates on this?

RodrigoRibeiroSantos:
> @RodrigoRibeiroSantos are there any updates on this?

Hi, we are finishing checking the removed districts. I will update the PR until this week.

RodrigoRibeiroSantos:
> @RodrigoRibeiroSantos are there any updates on this?

@jloutsenhizer Hi, I just updated the districts removed. Thanks

identifiers/country-mx/districts.csv
jloutsenhizer: This is very close, I think the only part that still needs fixed is the new districts still have the `2024-06-02` date for both `validFrom` and `validThrough`. Those ones just need to be updated so the `2024-06-02` date is only set for the `validFrom` date.

jloutsenhizer:
I left one more comment to correct around `validThrough` on the new districts. Otherwise this is close. Once you fix that make sure you also run `complie.py` again so that `country-mx.csv` gets updated.

@jpmckinney FYI for second review

jpmckinney:
I reviewed, and can approve once the error is corrected.

identifiers/country-mx.csv
```diff
@@ -1,4 +1,4 @@
-id,name
+id,name,validFrom,validTo
```
jpmckinney, line 1:
```suggestion
id,name,validFrom,validThrough

```

jpmckinney: Maybe just run `./scripts/compile.py mx`

RodrigoRibeiroSantos: Runned compile.py mx, that rewrite the file country-mx. If needs any changes, let me know.

image

jloutsenhizer (approved):
This PR LGTM

jpmckinney (approved):

PR #360, opened 2023-12-13
India - Assam Lok Sabha OCDID Update

mayanksahu123:
Hi Zohour,

This Pull Request contains below updated files for India - Assam Lok Sabha,

1. Added Valid To dates to kaliabor & kaliabor in lok sabha.csv
2. Added new constituency kaliabor & sonitpur in lok sabha.csv
3. Renamed mangaldoi, gauhati, autonomous_district, karimganj, silchar, nowgong constituencies by adding new ocd id in in lok sabha.csv & also added them in alias.csv


identifiers/country-in/lok_sabha.csv
jloutsenhizer, line 1: Please sort the lines by OCD ID to minimize diffs

jloutsenhizer (changes requested):
Do you have a source you can share for the new districts? I went off of what was available on [this Wikipedia article](https://en.wikipedia.org/wiki/List_of_constituencies_of_the_Lok_Sabha#Assam_(14)) and based on that all of the district changes in Assam are renames, but based on this PR some are renames and some are replacements. Could you provide some clarification?

identifiers/country-in/lok_sabha.csv
mayanksahu123: @jloutsenhizer 

Earlier Kaliabor and Tezpur were separate constituencies. But due to recent restructuring, Kaliabor comes under group of districts which fall under Kaziranga(new Lok Sabha Constituency) and similarly Tezpur has moved under the group of districts under Sonitpur(new Lok Sabha Constituency). For more clarification please check the attached snip from official document by ECI.
 
Snip

So these are the cases of restructuring not renaming. For your reference we are attaching official document and ECI link too underneath

https://www.eci.gov.in/files/file/15220-delimitation-of-parliamentary-and-assembly-constituencies-in-state-of-assam-%E2%80%93-final-notification-%E2%80%93-regarding/

[Final Order and Notification.pdf](https://github.com/opencivicdata/ocd-division-ids/files/13724831/Final.Order.and.Notification.pdf)

also we have sorted lines in lok_sabha.csv by OCD ID.

jloutsenhizer: I'm not sure I follow the logic, Under Kaziranga is Kaliabor the assembly district and under Sonitpur is Tezpur the assembly district. But when I take a look at the previous districts from https://www.eci.gov.in/files/file/15220-delimitation-of-parliamentary-and-assembly-constituencies-in-state-of-assam-%E2%80%93-final-notification-%E2%80%93-regarding/

They look very similar. For example, when i compare the old Tezapur Parliamentary Constituency and the new Sonitpur Parliamentary constituency I see one assembly district changed:

| old Tezapur  | new Sonitpur |
| ------------------ | -------------------- |
| Dhekiajuli | Dhekiajuli |
| Barchalla | Barchalla |
| Tezpur | Tezpur |
| Rangapara | Rangapara |
| Gohpur | Gohpur |
| Biswanath | Biswanath |
| Bihpuria | Bihpuria |
| Behali | Behali (SC) |
| **Sootea** | **Naduar** |

Yet a change that's considered a rename has significantly more changes in the assembly districts:

| old Mangaldoi | new Darrang-Udalguri |
| - | - |
| Kamalpur | Kamalpur |
|  Rangiya| Rangia |
| Sipajhar | Sipajhra |
| Dalgaon | Dalgaon |
| Udalguri (ST) | Udalguri (ST) |
| Majbat | Mazbat |
| **Mangaldoi (SC)** | **Tamulpur (ST)** |
| **Kalaigaon** | **Goreswar** |
| **Panery** | **Bhergaon** |
| **Nalbari** | **Tangla** |
|  | **Mangaldai** |

Do specific individual assembly districts carry the identity of the Parliament Constituency they are a part of? 

A rule of thumb I've generally used when deciding if a district is a rename or a new district is whether or not it'd be incorrect to consider an officeholder of this district to have a continuous tenure representing the same place. So for example,

Suppose whoever holds Mangaldoi was elected in 2019 and is elected for Darrang-Udalguri in 2024, would it be correct to say that they've represented Darrang-Udalguri since 2019?

If we substitute Tezapur and Sonitpur into that example, is it incorrect to say the officeholder has represented Sonitpur since 2019?

If both of those cases are correct then modeling both as renames makes sense to me. Otherwise they should be modeled as new districts. My intuition would be to treat each new name as a new district, but I'm not familiar enough to assert that with certainty.

jloutsenhizer: Thanks, these changes look good to me.

jloutsenhizer (approved):

identifiers/country-in/lok_sabha.csv
```diff
@@ -28,20 +28,27 @@ ocd-division/country:in/state:ap/cd:visakhapatnam,Andhra Pradesh Lok Sabha const
...
+ocd-division/country:in/state:as/cd:kaziranga,Assam Lok Sabha constituency Kaziranga,2024,
 ocd-division/country:in/state:as/cd:kokrajhar,Assam Lok Sabha constituency Kokrajhar (ST),,
 ocd-division/country:in/state:as/cd:lakhimpur,Assam Lok Sabha constituency Lakhimpur,,
-ocd-division/country:in/state:as/cd:mangaldoi,Assam Lok Sabha constituency Mangaldoi,,
+ocd-division/country:in/state:as/cd:mangaldoi,Assam Lok Sabha constituency Mangaldoi,,2024
+ocd-division/country:in/state:as/cd:nagaon,Assam Lok Sabha constituency Nagaon,,
 ocd-division/country:in/state:as/cd:nowgong,Assam Lok Sabha constituency Nowgong,,
 ocd-division/country:in/state:as/cd:silchar,Assam Lok Sabha constituency Silchar,,
-ocd-division/country:in/state:as/cd:tezpur,Assam Lok Sabha constituency Tezpur,,
+ocd-division/country:in/state:as/cd:silchar_sc,Assam Lok Sabha constituency Silchar (SC),,
```
jpmckinney, line 49: Hmm, I think we shouldn't be adding silchar_sc. We don't add the (SC) as part of the ID in other cases. (We shouldn't add the alias either, if we remove it here.)

jpmckinney (commented):
The source links are returning 404 for me.

It looks like these are ending:

* Autonomous District
* Kaliabor
* Mangaldoi
* Tezpur

And these are new:

* Darrang-Udalgiri
* Diphu
* Karizanga
* Sonitpur

Then gauhati and howgong are being renamed. We should remove those two from lok_sabha and leave them only in aliases.

identifiers/country-in/lok_sabha.csv
mayanksahu123: HI @jpmckinney  We have removed gauhati and nowgong from loksabha.csv & removed silchar_sc from both loksabha.csv and from alias.csv & have updated this changes in country-in.csv.

jpmckinney (approved):

raghu1061998:
Hi Team, please see below points regarding the lok_sabha.csv

- Name change: - The name of Anantnag constituency is changed to Anantnag-Rajouri. The current OCD ID ‘ocd-division/country:in/territory:jk/cd:Anantnag’ shows the name as ‘Jammu and Kashmir Lok Sabha constituency Anantnag’ in the repository, need to be changed to ‘Jammu and Kashmir Lok Sabha constituency Anantnag-Rajouri’. 

- Ladakh: - The present OCD ID of Ladakh Constituency ‘ocd-division/country:in/territory:jk/cd:ladakh’ shows the name as ‘Jammu and Kashmir Lok Sabha constituency Ladakh’.  Ladakh is a separate UT now and hence, it should be ‘Ladakh Lok Sabha constituency Ladakh’. (Looks like the territory was updated but it still remains under JK instead of LA)

PR #361, opened 2023-12-14
EU-Parliament Constituencies for Irleand

sguenther85:
added EU-Parliament Constituencies for Irleand and also for the EU election 2024 in Irleand. 
These are the first constituencies which are not the same like for the lower house of the country. 
So the whole thing is also open for discussion around the right type for the paths. I used "eu" right now, because the type described it the best for which usecase they are. But let me know if you have better suggestions.

### DIFF
--- a/identifiers/country-ie.csv
+++ b/identifiers/country-ie.csv
@@ -39,6 +39,6 @@ ocd-division/country:ie/ed:tipperary,Tipperary
 ocd-division/country:ie/ed:waterford,Waterford
 ocd-division/country:ie/ed:wexford,Wexford
 ocd-division/country:ie/ed:wicklow,Wicklow
-ocd-division/country:ie/eu:dublin,Dublin
-ocd-division/country:ie/eu:midlands-north-west,Midlands-North-West
-ocd-division/country:ie/eu:south,South
+ocd-division/country:ie/eup:dublin,Dublin
+ocd-division/country:ie/eup:midlands-north-west,Midlands-North-West
+ocd-division/country:ie/eup:south,South
--- a/identifiers/country-ie.csv
+++ b/identifiers/country-ie.csv
@@ -39,3 +39,6 @@ ocd-division/country:ie/ed:tipperary,Tipperary
 ocd-division/country:ie/ed:waterford,Waterford
 ocd-division/country:ie/ed:wexford,Wexford
 ocd-division/country:ie/ed:wicklow,Wicklow
+ocd-division/country:ie/eu:dublin,Dublin
+ocd-division/country:ie/eu:midlands-north-west,Midlands-North-West
+ocd-division/country:ie/eu:south,South
--- a/identifiers/country-ie/README.md
+++ b/identifiers/country-ie/README.md
@@ -1,6 +1,11 @@
-# OCD Division Identifier
+# Open Civic Data Divisions: Ireland
 
 This folder list the following administrative divisions of Ireland:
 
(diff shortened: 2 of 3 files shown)


sguenther85:
@azuser @jloutsenhizer Can anybody have a first look. Would be great if we can clarify this year how we want to deal with extra eu election constituencies of a country.
We have after that three other countries to add for the eu election.

identifiers/country-ie/README.md
```diff
@@ -0,0 +1,6 @@
+# OCD Division Identifier
+
```
HKSenior, line 2: Can you add a `Types` section listing out the various types used in this folder. For eg: https://github.com/opencivicdata/ocd-division-ids/blob/master/scripts/country-ca/README.md?plain=1#L23

identifiers/country-ie/README.md
```diff
@@ -0,0 +1,6 @@
...
+This folder list the following administrative divisions of Ireland:
+
+* Lower House Constituencies [ref](https://en.wikipedia.org/wiki/D%C3%A1il_constituencies). Used type: **ed**.
```
HKSenior, line 5:
```suggestion
* Lower House Constituencies [ref](https://en.wikipedia.org/wiki/D%C3%A1il_constituencies).
```

You can remove this (here and below) once the `Type` section is added above.

identifiers/country-ie.csv
```diff
@@ -39,3 +39,6 @@ ocd-division/country:ie/ed:tipperary,Tipperary
 ocd-division/country:ie/ed:waterford,Waterford
 ocd-division/country:ie/ed:wexford,Wexford
 ocd-division/country:ie/ed:wicklow,Wicklow
+ocd-division/country:ie/eu:dublin,Dublin
```
HKSenior, line 42: I'd recommend using `eup` (for EU Parliament) instead of `eu` for the type.

HKSenior (changes requested):

jloutsenhizer:
+1 to @HKSenior's suggestions, otherwise this PR looks good to me.

identifiers/country-ie/README.md
```diff
@@ -0,0 +1,6 @@
+# OCD Division Identifier
```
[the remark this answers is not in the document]
sguenther85: Followed your suggestion. Send an update. Please have a look.

identifiers/country-ie/README.md
sguenther85: Followed your suggestion. Send an update. Please have a look.

identifiers/country-ie/README.md
sguenther85: Followed your suggestion. Send an update. Please have a look.

identifiers/country-ie.csv
sguenther85: Followed your suggestion. Send an update. Please have a look.

HKSenior (approved):

sguenther85:
Hey together, please wait with the merge. 
There is a discussion on DPA/Google side where we have to add the constituencies for the eup election.
This way, or in the country-eu.csv with all other constituencies which belongs the the eup election.

sguenther85:
Google/DPA: Sorry, but we discussed this internally and the EUP constituencies should be in the eu-countries file. Basically all of the districts for the EU Parliament (those that have jurisdiction of the EU) should be in one file.

So i will close this pull request and create a new one with all constituencies which will be used for the eup election.

PR #364, opened 2024-01-25
Canada electoral districts 2023

NikolasYo:
Adding the constituency changes according to the Canadian electoral commission (2023)

Also attaching a README file with sources for those changes


evannjw (changes requested):
Can you add `2024-04-23` as validThrough date to the ids in the identifiers/country-ca/ca_federal_electoral_districts-2013.csv file?

You need to run `./scripts/compile.py ca` to compile changes into identifiers/country-ca.csv

Also consider adding the 2013 ids which share the same district name as aliases here. Will defer to @jpmckinney on this.

jpmckinney:
@NikolasYo Can you review #327 first? Ideally we would merge that, and then make additional changes.

NikolasYo:
We reviewed the previous request #327 with @zikowang. From our side, you can merge, and we will work on the new EDs afterwards. We will keep the same ID for the EDs that stay the same and add validFrom and validThrough dates.

jpmckinney:
@NikolasYo Do you mean you would like this PR to be merged, but that you'll later make aliases for these to other OCDIDs (following the pattern in #327)?

zikowang:
@jpmckinney we would wait for #327 to be merged and adjust this PR after accordingly.

zikowang:
The 2013 file is updated (added `validThrough`).
The `aliases.csv` is updated with 2023 aliases matching the 2013 aliases.
The README updated with sources for the 2023 changes.

@evannjw @jpmckinney would be great if you could have a look into this approach and if am missing something.

identifiers/country-ca/aliases.csv
```diff
@@ -1,4 +1,237 @@
...
+ocd-division/country:ca/ed:59035-2023,ocd-division/country:ca/province:bc/ed:vancouver_centre
+ocd-division/country:ca/ed:59036-2023,ocd-division/country:ca/province:bc/ed:vancouver_east
+ocd-division/country:ca/ed:59038-2023,ocd-division/country:ca/province:bc/ed:vancouver_granville
+ocd-division/country:ca/ed:59039-2023,ocd-division/country:ca/province:bc/ed:vancouver_kingsway-1997
+ocd-division/country:ca/ed:59040-2023,ocd-division/country:ca/province:bc/ed:vancouver_quadra
+ocd-division/country:ca/ed:59042-2023,ocd-division/country:ca/province:bc/ed:victoria-1925
+ocd-division/country:ca/ed:59043-2023,ocd-division/country:ca/province:bc/ed:west_vancouver-sunshine_coast-sea_to_sky_country
+ocd-division/country:ca/ed:60001-2023,ocd-division/country:ca/territory:yt/ed:yukon-1953
+ocd-division/country:ca/ed:61001-2023,ocd-division/country:ca/territory:nt/ed:northwest_territories-2015
+ocd-division/country:ca/ed:62001-2023,ocd-division/country:ca/territory:nu/ed:nunavut
```
jpmckinney, line 234: Looks good, but why do only 233 out of 343 "-2023" OCD IDs have aliases?

zikowang: The OCDs that are new or habv other name in the 2023 file compared to the 2013 file don't have aliases.

Should there be an alias for each entry in the 2023 file?

jpmckinney: Yes, in #323 we basically agreed to have:

1. ID and year based OCD IDs, for easy reconciliation with new boundary files.
2. Name-based OCD IDs that reconcile with data from the Library of Parliament.

The `scripts/country-ca/ca_federal_electoral_districts.py` script was created to assist here. You might need to run it (and possibly edit it to accommodate the new IDs).

NikolasYo:
The source mentioned https://lop.parl.ca/sites/ParlInfo/default/en_CA/ElectionsRidings/Ridings is not up to date as it contains 338 active constituencies instead of 343

After examining the changes in the electoral districts we came across multiple examples of electoral district names staying the same, despite changes in their borders. You can see it visually here https://redecoupage-redistribution-2022.ca/ebv2/index.html?locale=en-ca .

For example ED Selkirk—Interlake—Eastman kept the same name after the 2023 changes, but it’s border was modified, losing an area in the south against ED Winnipeg West and gaining an area in the north from ED Churchill—Keewatinook Aski. Such changes (Same ED name but different borders) are noticeable in many constituencies.

Our proposal would be to treat all the 2023 Electoral Districts as new, due to the numerous changes in the borders. Would you prefer to treat constituencies that have changed their borders as continuous solely based on the fact of them keeping the same name (Alias)?

jpmckinney:
> Our proposal would be to treat all the 2023 Electoral Districts as new, due to the numerous changes in the borders. 

A division (political geography) is not the same as a boundary per [OCDEP 2](https://github.com/opencivicdata/docs.opencivicdata.org/blob/master/proposals/0002.rst). So, borders can change without yielding a new division.

> Would you prefer to treat constituencies that have changed their borders as continuous solely based on the fact of them keeping the same name (Alias)?

Name doesn't matter at all in determining continuity over time.

* In Canada, Members of Parliament frequently rename districts for any reason, without any boundary change.
* On the flip side, there are cases where the Library of Parliament deems two divisions to be distinct, despite having the same name, due to perhaps some major change in boundary.

To answer the question behind your question: Whether two divisions are different or the same is a *political decision*. In Canada, we treat the Library of Parliament as the authoritative source on what has been politically decided.

LOP has not yet caught up on recent changes. So, for now, we can just create the "-2023" OCDIDs and NOT add any aliases, until our source for aliases is available.

zikowang:
@jpmckinney thanks for the clarification.
I pushed a new commit removing the aliases for CA 2023 ocd-ids.

jpmckinney (approved):

zikowang:
@jloutsenhizer @azuser 
could you please have a short look?

evannjw (approved):

jloutsenhizer (approved):

PR #365, opened 2024-02-08
323 canada fed rename

jpmckinney:


jloutsenhizer:
If `identifiers/country-ca.csv` is the only file with conflicts, I think running compile.py and replacing the file would be easier than going through the actual conflicts

jloutsenhizer (approved):

PR #369, opened 2024-03-19
South Korea single_member_constituencies update

MarcinMisztalTR:
South Korea single_member_constituencies update

Hello,
South Korea confirmed its new electoral districts for the April 10, 2024 elections
We have added the new districts with new OCD-IDs
Until the winning candidates take office on 5/30/2024, we will need to have both old and new districts in place. 
This is because existing officeholders will still be assigned to the "old" OCD-IDs until new officeholders take office in the new districts. 

List of new zones (validFrom set to 05/30/2024):
Nam District First Busan
Buk District First
Buk District Second
Gangseo District
Dong District and Gunwi County First
Dong District and Gunwi County Second
Seo District Third
Pyeongtaek City Third
Dongducheon City, Yangju City and Yeoncheon County First
Dongducheon City, Yangju City and Yeoncheon County Second
Ansan City First
Ansan City Second
Ansan City Third
Hanam City First
Hanam City Second
Hwaseong City FOURTH
Gunsan City, Gimje City and Buan County First
Gunsan City, Gimje City and Buan County Second
Namwon City, Jangsu County, Imsil County and Sunchang County
Wanju County, Jinan County and Muju County

List of zones which will be obsolete (validThrough set to 05/29/2024):
Ansan City Danwon District First
Ansan City Danwon District Second
Ansan City Sangnok District First
Ansan City Sangnok District Second
Bucheon City Fourth
Dongducheon City and Yeoncheon County
Gimje City and Buan County
Gunsan City
Hanam City
Namwon City, Imsil County and Sunchang County
Nowon III
Wanju County, Jinan County, Muju County and Jangsu County
Yangju City

country-kr.csv file was updated by compile script.


azuser:
The valid from for new constituencies should be set to the date of the election for which they will be used: April 10th

azuser (changes requested):
Added few comments to resolve.

+ I was thinking that some of these new district are renamed not newly created. Can you double check ?

MarcinMisztalTR:
Hello,
@azuser thank you for the review, could you please review it again?
Thanks

azuser:
I would expect 254 single seat constituencies. In the change, only 253 will be valid starting 2024-04-10. Can you double check the numbers and provide the sources if needed ?

MarcinMisztalTR:
@azuser Sorry for the inconvenience, missing item recovered, now should be 254

jloutsenhizer (approved):
These changes look good to me

jpmckinney:
* Where are the names sourced from?
  * Use sentence case for FOURTH.
  * Parenthesize (Busan) consistently.
* How are IDs derived from names?
  * "district_" is inserted into some that were previously just "City First", etc. The names have City, County, District, or nothing at all (Dobong First, etc.). Not sure why "district_" is added if it's not in the name.
* It looks like we're removing 20 and adding 19. Is it correct that the number of districts before/after changes by 1?

Noting that these were also obsoleted (not listed above)

Buk District and Gangseo District First (Busan)
Buk District and Gangseo District Second (Busan)
Nam District First (Busan)
Nam District Second (Busan)
Dong District First (Daegu)
Dong District Second (Daegu)

About dates: I think we have always had them either one day apart or the same day (probably more correct and easier on users to have one day apart). I don't think it's correct for there to be overlap, as that confuses two different perspectives. If the question is "what is the validity, with respect to the division to be represented by the winner of an election?" then the date is 2024-04-10. If the question is "what is the validity, with respect to the division represented by the active officeholder?" then the date is 2024-05-29.

To date, we have only concerned ourselves with the election date. That said, in pretty much every election, the officeholder remains in office for some time after the election date (notably, because many elections are not 100% decided on the election date).

If we want to answer both questions, we should have different fields. However, I don't know if any users of OCDIDs really care to know the date when responsibilities are transferred. This is also the sort of date that sometimes moves if the election is hard to decide – and whereas I think we try to have the correct election dates (which also sometimes move), I don't think anyone is really monitoring whether this other date moves.

I would assume any current user of OCDIDs is just using the logic of: is today between validFrom and validThrough? Then the division is relevant. But this simple logic will fail with the dates in this PR. Users would now have to ask, is today after validFrom, and – if the division sets a validThrough in the future – is today also before the earliest future validFrom among other divisions? (This logic is too bizarre.)

MarcinMisztalTR:
> * Where are the names sourced from?
>   
>   * Use sentence case for FOURTH.

Fixed. (changed to Fourth)

>   * Parenthesize (Busan) consistently.
> * How are IDs derived from names?
>   
>   * "district_" is inserted into some that were previously just "City First", etc. The names have City, County, District, or nothing at all (Dobong First, etc.). Not sure why "district_" is added if it's not in the name.
> * It looks like we're removing 20 and adding 19. Is it correct that the number of districts before/after changes by 1?
> 
> Noting that these were also obsoleted (not listed above)
> 
> Buk District and Gangseo District First (Busan) Buk District and Gangseo District Second (Busan) Nam District First (Busan) Nam District Second (Busan) Dong District First (Daegu) Dong District Second (Daegu)
> 

These are sourced from the official list. The same one in Korean we have supplied and translated.
We were trying not to change every district. These have been created over many years and not always with the same syntax.
If you want us to go through and change them so they are all consistent, we can do that but it mean many more changes and will take at least 2 weeks to do.
The correct number of districts is 254 and that is what we have provided. We may have had a district split into two unnecessarily but I do not want to go back and review all the historical ones that are now out of date.
Please let us know if you would prefer us to redo all of the current districts so the syntax is consistent.

> About dates: I think we have always had them either one day apart or the same day (probably more correct and easier on users to have one day apart). I don't think it's correct for there to be overlap, as that confuses two different perspectives. If the question is "what is the validity, with respect to the division to be represented by the winner of an election?" then the date is 2024-04-10. If the question is "what is the validity, with respect to the division represented by the active officeholder?" then the date is 2024-05-29.
> 
> To date, we have only concerned ourselves with the election date. That said, in pretty much every election, the officeholder remains in office for some time after the election date (notably, because many elections are not 100% decided on the election date).
> 
> If we want to answer both questions, we should have different fields. However, I don't know if any users of OCDIDs really care to know the date when responsibilities are transferred. This is also the sort of date that sometimes moves if the election is hard to decide – and whereas I think we try to have the correct election dates (which also sometimes move), I don't think anyone is really monitoring whether this other date moves.
> 
> I would assume any current user of OCDIDs is just using the logic of: is today between validFrom and validThrough? Then the division is relevant. But this simple logic will fail with the dates in this PR. Users would now have to ask, is today after validFrom, and – if the division sets a validThrough in the future – is today also before the earliest future validFrom among other divisions, minus one day? (This logic is too bizarre.)

We can adjust the zone dates for To and From to equal the election date if that's easiest. 
But the "old" districts are valid until the new officeholder takes office. That is why the date is 5/29/24.
So given the above, will we get an error if we have an officeholder with an end date of 5/29/24 (This date is stipulated in CDF guidelines) but their district is only valid through April 10,2024?

jpmckinney:
>> Where are the names sourced from?

You didn't answer. Please provide a URL.

> These are sourced from the official list. The same one in Korean we have supplied and translated. We were trying not to change every district. These have been created over many years and not always with the same syntax. If you want us to go through and change them so they are all consistent, we can do that but it mean many more changes and will take at least 2 weeks to do. The correct number of districts is 254 and that is what we have provided. We may have had a district split into two unnecessarily but I do not want to go back and review all the historical ones that are now out of date. Please let us know if you would prefer us to redo all of the current districts so the syntax is consistent.

I'm confused. Let's take a specific example:

ocd-division/country:kr/ed:ansan_city_district_1,Ansan City First,,2024-04-10

This is one of the new districts. There is no reason for the ID to contain "district_". The name you provide above and in the PR ("Ansan City First") doesn't contain "District". The name is what comes from the official sources. The only change would be to remove "district_" in this case (and the other similar cases). This simple transformation does not take 2 weeks.

>> Parenthesize (Busan) consistently.

This change only requires "Nam District First Busan" -> "Nam District First (Busan)". This is one of the new districts. In every other case, a name that appears after First, Second, etc. is parenthesized. There is no reason to have this one exception.

> We can adjust the zone dates for To and From to equal the election date if that's easiest.

What we have done in other countries is set validThrough to the day **before** the election, and validFrom to the same day as the election.

> But the "old" districts are valid until the new officeholder takes office. That is why the date is 5/29/24.

As I already explained, this is true in pretty much every country. The problem is that you are trying to use the same fields to answer two different questions.

> So given the above, will we get an error if we have an officeholder with an end date of 5/29/24 (This date is stipulated in CDF guidelines) but their district is only valid through April 10,2024?

I don't know what CDF guidelines are - you should provide links when referring to sources.

As I explained, the fields are used to answer a specific question. Let's say a voter uses a geolocation service (that is part of a voter information service) to determine which district they fall within that is valid as of election day (2024-04-10). In the current PR, where districts have changed, like Ansan, voters will get two answers: the old district (validThrough 2024-05-29) and the new district (2024-04-10). This is clearly undesirable. The voter information service should be responding with a single district: the district to which the new representatives are being elected.

 If you ask a different question, like "which valid district am I currently within, and who is the current officeholder?" – then, of course, you will need different dates. But if that's what you want, then add new fields for those dates. Don't use fields that are designed to answer a different question – that will not work.

MarcinMisztalTR:
Hello,
according to review:
- validThrough date changed from 29/05/2024 to 9/4/2024
- removed _district for some districts (if no district in the name)
- changed Busan to (Busan)

- added 2 new districts:
Yeongju City, Yeongyang County and Bonghwa County
Uiseong County, Cheongsong County, Yeongdeok County, Uljin County

jpmckinney:
Thank you. Last thing is: Can you provide a URL or document for the source of these names?

There are 21 districts expiring on 2024-04-09 and 22 districts starting on 2024-04-10. It seems a bit strange that there is such a small change in the number of districts, so I'd like to compare against a source.

MarcinMisztalTR:
Hello,
This is the document/source:
[22nd National Assembly election electoral districts.pdf](https://github.com/opencivicdata/ocd-division-ids/files/15296017/22nd.National.Assembly.election.electoral.districts.pdf)

jpmckinney (approved):
Looks good and the new/surviving districts add up to 254 like in the source.

PR #373, opened 2024-04-07
Changes EUP poland constituencies

sguenther85:
We have noticed that Poland has changed its constituencies again. Not the number, but the whole naming of them. These also no longer match 2019, which was the only information we had at the time.

https://wybory.gov.pl/pe2024/pl/okregi

### DIFF
--- a/identifiers/country-eu.csv
+++ b/identifiers/country-eu.csv
@@ -32,19 +32,19 @@ ocd-division/country:lv,Latvia
 ocd-division/country:mt,Malta
 ocd-division/country:nl,Netherlands
 ocd-division/country:pl,Poland
-ocd-division/country:pl/eup:bydgoszcz,No. 2 - Bydgoszcz
-ocd-division/country:pl/eup:gdańsk,No. 1 - Gdańsk
-ocd-division/country:pl/eup:gorzów_wielkopolski,No. 13 - Gorzów Wielkopolski
-ocd-division/country:pl/eup:katowice,No. 11 - Katowice
-ocd-division/country:pl/eup:kraków,No. 10 - Kraków
-ocd-division/country:pl/eup:lublin,No. 8 - Lublin
-ocd-division/country:pl/eup:olsztyn,No. 3 - Olsztyn
-ocd-division/country:pl/eup:poznań,No. 7 - Poznań
-ocd-division/country:pl/eup:rzeszów,No. 9 - Rzeszów
-ocd-division/country:pl/eup:warszawaw_i,No. 4 - Warszawa
-ocd-division/country:pl/eup:warszawaw_ii,No. 5 - Warszawa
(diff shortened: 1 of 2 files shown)


sguenther85:
@jloutsenhizer

jloutsenhizer:
It's interesting that the Official Polish websites have switched how they're referring to these constituencies. Also interestingly, this appears to be more in line with how Offical EU parliament sources referred to them as well (2004 election source): https://www.europarl.europa.eu/elections2004/ep-election/sites/en/yourvoice/pl/law.html

It might make sense for these to be aliases for the originally submitted names to be aliases for the udpated names since they appear to have both been used to refer to these same districts in the past, but I'll leave that decision to your judgement.

jloutsenhizer (approved):

sguenther85:
> It's interesting that the Official Polish websites have switched how they're referring to these constituencies. Also interestingly, this appears to be more in line with how Offical EU parliament sources referred to them as well (2004 election source): https://www.europarl.europa.eu/elections2004/ep-election/sites/en/yourvoice/pl/law.html
> 
> It might make sense for these to be aliases for the originally submitted names to be aliases for the udpated names since they appear to have both been used to refer to these same districts in the past, but I'll leave that decision to your judgement.

I believe that an "alias" is not necessary in this case because we have not yet used it to represent an election here and these were also submitted for the first time 3 months ago for exactly this intended election.

jpmckinney:
Should we use the number after Nr in the ID? The IDs right now are kind of meaningless anyway, so might as well abbreviate.

sguenther85:
@jpmckinney 
> Should we use the number after Nr in the ID? The IDs right now are kind of meaningless anyway, so might as well abbreviate.

i think that would have made problems for us here, because the numbers are still the same. 
That means we would have had the same numbers, but different names. That would then be difficult to represent historically in the ocd file, wouldn't it?

But I basically agree with you that the current ids are not particularly nice, but I haven't found a useful alternative

jpmckinney:
In general, a name can change without changing the identity of the political division.

Are the boundaries the same, or is basically nothing the same since 2019, other than the reuse of the numbers 1-13?

jpmckinney:
Okay, then I assume their identity is the same, and it's just the name that changed (in Canada, Members of Parliament love to rename their electoral district – it's still the same district).

So, I would prefer to use the numbers in the ID.

jloutsenhizer:
This seems like a fair use of numbers, it's unclear how often the districts will change, but if they do redistrict and reuse the numbers for totally different districts we can take a similar approach that we did for Canada where we attach a year suffix to the IDs to give us unique ones

sguenther85:
@jpmckinney @jloutsenhizer done

jpmckinney (approved):

PR #374, opened 2024-04-26
Add in Nunavut districts

seamuslee001:
This is based on the list of Constituencies here https://www.elections.nu.ca/en/constituency-map


evannjw (commented):
Since these districts don't seem to have ids, consider using names as district identifiers. ex. `ocd-division/country:ca/territory:nu/ed:1` -> `ocd-division/country:ca/territory:nu/ed:aggu`

jpmckinney:
Agreed

jpmckinney (approved):

PR #380, opened 2024-05-15
fix(ca): Disambiguate federal districts from provincial districts using fed type

jpmckinney:
In https://github.com/opencivicdata/ocd-division-ids/issues/323#issuecomment-1300688869 @evannjw and I agreed add the province or territory to the ID for the federal district, but this of course clashes with provincial and territorial districts, unless we disambiguate with a different type ID.


evannjw (approved):

identifiers/country-ca.csv
```diff
@@ -10775,490 +10780,490 @@ ocd-division/country:ca/province:on/ed:98,Welland,,,,,,,,,,,,,,
...
-ocd-division/country:ca/province:on/ed:mississauga_west,Mississauga West,,2004-05-22,,,,,,,,,,,,1988-10-01
-ocd-division/country:ca/province:on/ed:monck,Monck,,1898-04-23,,,,,,,,,,,,1867-08-06
-ocd-division/country:ca/province:on/ed:muskoka,Muskoka,,1882-05-16,,,,,,,,,,,,1872-07-08
-ocd-division/country:ca/province:on/ed:muskoka-1904,Muskoka,,1925-09-04,,,,,,,,,,,,1904-09-20
-ocd-division/country:ca/province:on/ed:muskoka-ontario,Muskoka—Ontario,,1949-04-29,,,,,,,,,,,,1925-09-05
-ocd-division/country:ca/province:on/ed:muskoka_and_parry_sound,Muskoka and Parry Sound,,1904-09-19,,,,,,,,,,,,1882-05-17
-ocd-division/country:ca/province:on/ed:nepean,Nepean,,1997-04-26,,,,,,,,,,,,1988-10-01
-ocd-division/country:ca/province:on/ed:nepean-2015,Nepean,,,,,,,,,,,,,,2015-08-02
-ocd-division/country:ca/province:on/ed:nepean-carleton,Nepean—Carleton,,1988-09-30,,,,,,,,,,,,1979-03-26
-ocd-division/country:ca/province:on/ed:nepean-carleton-1997,Nepean—Carleton,,2015-08-01,,,,,,,,,,,,1997-04-27
```
jloutsenhizer, line 11052: With this change most of the federal district OCD IDs that used `ed` prefix are completely dropped. I think it would be best to preserve them with `sameAs` aliases so that they map to their new `fed` counterpart.

jloutsenhizer (changes requested):

identifiers/country-ca.csv
jpmckinney: I don't think your suggestion is workable. #328 BROKE all provinces and territories by adding federal "ed" OCDIDs in the same space occupied by provincial "ed" OCDIDs. We cannot keep federal "ed" OCDIDs in that space. The only solution is to essentially revert and change to "fed". If we "sameAs" we just perpetuate the issue, because "sameAs" OCDIDs exist in the same space as non "sameAs" OCDIDs.

jpmckinney: The federal "ed" districts that hang directly off "country:ca" are untouched.

jloutsenhizer: To clarify is the issue that there are current collisions or that consumers of OCD IDs can no longer distinguish federal CA districts from provincial ones based solely on the OCD ID?

Either way I think it's fair to do this for the latter case, but I'd like to understand the issue better. 

I still have some concerns about systems that rely on OCD IDs not completely disappearing or later changing meaning. An alternate option for dealing with this could be to allow OCD IDs to be marked as invalid/deprecated. However I don't think we have an existing mechanism for that so it'd require an OCDEP.

jpmckinney: It’s at least the latter, and I think the former is also true.

Since the affected IDs haven’t been around that long (and since they weren’t that usable), I would just delete them.

Another option besides deprecation is versioning (also requires OCDEP / governance discussion). Current systems wouldn’t be aware of deprecation semantics, so they’d treat the deprecated OCDIDs as valid.

jloutsenhizer: Thanks, I think we can continue the discussion of how to handle situations like this in an issue. I'm ok with moving forward with this PR.

PR #381, opened 2024-05-17
Update division names for Edmonton, CA

rafe-murray:
This PR updates the division names to reflect Edmonton's renaming of its wards


identifiers/country-ca/ca_municipal_subdivisions.csv
```diff
@@ -2982,18 +2982,18 @@ ocd-division/country:ca/csd:4811052/ward:5,Ward 5
...
-ocd-division/country:ca/csd:4811061/ward:9,Ward 9
-ocd-division/country:ca/csd:4811061/ward:10,Ward 10
-ocd-division/country:ca/csd:4811061/ward:11,Ward 11
-ocd-division/country:ca/csd:4811061/ward:12,Ward 12
+ocd-division/country:ca/csd:4811061/ward:anirniq,Anirniq
+ocd-division/country:ca/csd:4811061/ward:dene,Dene
+ocd-division/country:ca/csd:4811061/ward:ipiihkoohkanipiaohtsi,Ipiihkoohkanipiaohtsi
+ocd-division/country:ca/csd:4811061/ward:karhiio,Karhiio
+ocd-division/country:ca/csd:4811061/ward:metis,Métis
+ocd-division/country:ca/csd:4811061/ward:nakota_sga,Nakota Isga
```
evannjw, line 2990: This looks to be missing a letter in the id, should be `ward:nakota_isga`

identifiers/country-ca.csv
```diff
@@ -7411,18 +7411,18 @@ ocd-division/country:ca/csd:4811052/ward:8,Ward 8,,,,,,,,,,,,,,
 ocd-division/country:ca/csd:4811056,Fort Saskatchewan,,,,,CY,N,Fort Saskatchewan,,City of Fort Saskatchewan,,,,,
 ocd-division/country:ca/csd:4811059,Sturgeon County,,,,,MD,Y,Sturgeon County,,Municipality of Sturgeon County,,,,,
 ocd-division/country:ca/csd:4811061,Edmonton,,,,,CY,Y,Edmonton,,City of Edmonton,,,,,
-ocd-division/country:ca/csd:4811061/ward:1,Ward 1,,,,,,,,,,,,,,
```
evannjw, line 7414: Consider keeping these and instead adding aliases in aliases.csv instead

seamuslee001: I was looking at the aliases file and seems to be more about federal divisions rather than council wards. @jpmckinney James any thoughts on if these should be shifted into the aliases file?

identifiers/country-ca/ca_municipal_subdivisions.csv
rafe-murray: Thanks! I fixed this in the newest commit

identifiers/country-ca.csv
jpmckinney: I think we need to validFrom/validThrough because the boundaries themselves have changed as well, with no clear mapping: https://edmonton.ctvnews.ca/edmonton-s-new-indigenous-ward-names-explained-1.5599364

I can update the PR.

jpmckinney:
I couldn't add commits to this PR so I opened https://github.com/opencivicdata/ocd-division-ids/pull/382

PR #386, opened 2024-06-07
Update Ottawa divisions

rafe-murray:
This PR update the division names of six of Ottawa's existing wards and adds one new ward in order to match the changes made in 2022 documented on [this page](https://ottawa.ca/en/city-hall/public-engagement/public-engagement-project-search/ottawa-ward-boundary-review-2020)


evannjw (commented):
Consider keeping the old district ids and adding a validThrough date on them then alias those ids to the new ones you're adding. Will defer to @jpmckinney here.

rafe-murray:
I tried doing this on my machine, but I got an error when I tried to run the compile script due to not actually changing the id's, just the names of the districts.

jpmckinney:
The approach for the municipal level is to update shapefiles in https://github.com/opennorth/represent-canada-data and then regenerate OCID-IDs based on that. Maintaining history at the municipal level is not really feasible.

https://open.ottawa.ca/datasets/ottawa::wards-2022-2026/explore?location=45.245954%2C-75.801100%2C0.96&showTable=true

jpmckinney (approved):
I updated the shapefile and they correspond to this PR.

PR #388, opened 2024-06-26
fix: Remove Kawartha Lakes Wards 9-16

rafe-murray:


jloutsenhizer (changes requested):
Did these never exist or did they exist but were abolished? If the latter then typically we'd set a `validThrough` date rather than completely remove them

rafe-murray:
They used to exist but they were removed in December 2018. I've added another commit with the validThrough dates. See https://www.kawarthalakes.ca/en/municipal-services/ward-boundaries.aspx

jloutsenhizer (approved):

PR #390, opened 2024-08-23
Add Denmark election districts OCD Ids

vald0506:
Adding OCDs for the three levels of Danish election jurisdictions.

See also new README.md (identifiers/country-dk/README.md) for details.

### DIFF
--- a/identifiers/country-dk.csv
+++ b/identifiers/country-dk.csv
@@ -1,2 +1,107 @@
-id,name
-ocd-division/country:dk,Denmark
+id,name,validFrom
+ocd-division/country:dk,Denmark,
+ocd-division/country:dk/election_province:hovedstaden,A. Hovedstaden,2007-01-01
+ocd-division/country:dk/election_province:sjaelland-syddanmark,B. Sjælland-Syddanmark,2007-01-01
+ocd-division/country:dk/election_province:midtjylland-nordjylland,C. Midtjylland-Nordjylland,2007-01-01
+ocd-division/country:dk/election_province:hovedstaden/ed:koebenhavns_storkreds,1. Københavns Storkreds,2007-01-01
+ocd-division/country:dk/election_province:hovedstaden/ed:koebenhavns_omegns_storkreds,2. Københavns Omegns Storkreds,2007-01-01
+ocd-division/country:dk/election_province:hovedstaden/ed:nordsjaellands_storkreds,3. Nordsjællands storkreds,2007-01-01
+ocd-division/country:dk/election_province:hovedstaden/ed:bornholms_storkreds,4. Bornholms Storkreds,2007-01-01
+ocd-division/country:dk/election_province:sjaelland-syddanmark/ed:sjaellands_storkreds,5. Sjællands Storkreds,2007-01-01
+ocd-division/country:dk/election_province:sjaelland-syddanmark/ed:fyns_storkreds,6. Fyns Storkreds,2007-01-01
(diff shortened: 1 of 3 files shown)


evannjw (changes requested):
The names of the districts have a list artifact (A., 1.) preceding them. Can you remove these?

jloutsenhizer (changes requested):
Overall LGTM, Denmark is divided into 3 election provicnes, which are then subdivided into 10 multimember constituencies and those are subdivided to 92 nominating districts which are all reflected.

One question, Greenland and Faroe Islands each elect  2 members of the Danish parliament and sounds like they could be subdivded further (from what I saw Greenland used to be). Have you considered representing these with their own OCD IDs?

vald0506:
Greenland, the Faroe Islands, and Denmark are all part of the Kingdom of Denmark. Both G and FI have wide-ranging autonomy with own parliaments and local governments (for which reason each country also hold own parliamentary and local elections), and neither are part of the EU (as D is). While G and FI have home rule in most domestic matters, D still controls foreign and security policy or currency policy (which also explains the need for representation in the Danish parliament with 2 seats each).

I'm new to this project so I may not be suited to decide whether or not G and FI should be included in the same folder as D. One solution could be to do as the case of Great Britain with a country divided into parts. Another could be to have G and FI have individual folders. As I see it, either way could work, but I don't if know there is precedent for doing (if any).

FYI, here are the consolidated acts of [Greenland](https://www.retsinformation.dk/eli/lta/2023/1167#:~:text=Fortegnelse%20over%20valgkredse%20i%20hver%20kommune) (columns: "Kommune", "Nomination district", "Polling station", and "Included towns") and [Faroe Islands](https://www.retsinformation.dk/eli/lta/2023/1432#:~:text=Bilag%201-,Fortegnelse%20over%20valgdistrikterne,-Syssel) (columns: "Syssel", "Included towns", and "Polling station") outlining the districts. In both (G and FI) cases, the subdivisions do not influence the distribution of seats as the calculations are based on the total number of votes in each country.

jloutsenhizer:
Thanks for the detailed explanation. It sounds like Greenland and Denmark both fall into fuzzy categories in terms of if they should have their own top-level `ocd-divison/country:` identifiers.

We don't necessarily need the top-level country identifier though, as these are seats on the parliament of Denmark it'd make sense to nest the OCD IDs for the electoral districts under `ocd-division/country:dk/`. This would match how oversea territories are handled for France, for example: https://github.com/opencivicdata/ocd-division-ids/blob/master/identifiers/country-fr/oversea.csv

jpmckinney:
I think rename election_province to electoral_province to match "electoral district" in the rest of this project.

I had to check whether we care about nomination districts: https://www.elections.im.dk/media/40794/the-electoral-system-in-denmark.pdf

> Nomination districts have no significance in terms of seat allocation between the political parties.

At first, this concept seemed similar to "polling divisions" in other countries. Polling divisions have no political consequence and are just a way of organizing the vote (e.g. asking people to go to a particular building to vote).

However, it looks like nomination districts are politically relevant in Denmark, since candidates can stand in a single district:

> There are two principal forms of list organisation for candidates: standing by district and standing in parallel:
> • In standing by district, each nomination district has its own candidate. The candidate is allotted all the votes
> cast for the party in the nomination district plus the personal votes cast for him or her in all the nomination
> districts of the multimember constituencies. The size of the nomination district, i.e. the number of voters, there-
> fore plays an important role.
> ...

PR #391, opened 2024-09-02
LT electoral district changes 2023-11-30

zikowang:
This PR adds the latest changes for Lithuania.

- added `validFrom` and `validThrough` columns
- added 4 new electoral districts
- invalidated 4 old electoral districts
- added a README with sources


jloutsenhizer (approved):
I've verfiied the 4 added districts show up int he provided source and the 4 removed districts do not.

One note, it's a bit odd we have the country-level ocd ID in the `electoral_districts.csv` file. Ideally that'd be in its own file. However, that's not caused by this change, so we can clean that up later.

jpmckinney (approved):
LGTM

PR #393, opened 2024-09-18
update Ireland electora districts

zikowang:
Adding the latest changes for Irelands lower house OCDs.

- added a `top_level` file for country-ie
- added `validFrom` and `validThrough` columns
- added 7 new electoral districts
- invalidated 3 old electoral districts
- added a README with sources


azuser:
What is the reason for using March 21st, 2025 as the effective date for the new and abolished constituencies?

NikolasYo:
> What is the reason for using March 21st, 2025 as the effective date for the new and abolished constituencies?

March 21st, 2025, is set as the effective date to ensure the current constituencies remain valid until the election on (most likely) March 22nd, 2025. The new constituencies take effect immediately after the election, aligning with the new legislative period. The "old" constituencies remain active for the current legislature.

azuser:
Since the upcoming election date is not yet defined, I would assume that the new constituencies can be valid even before March 21st, 2025. Right ?

NikolasYo:
Hello, the new constituencies can't be valid before the start of the new legislature, as the "old" ones will still be used by active Officeholders. The new constituencies will be valid in the upcoming election and upcoming legislature. The most likely date for the election is March 22nd, 2025. I suggest that we use March 21st as valid through and valid from date for now, and change it if the election happens on a different date. What do you think?

azuser:
> Hello, the new constituencies can't be valid before the start of the new legislature, as the "old" ones will still be used by active Officeholders. The new constituencies will be valid in the upcoming election and upcoming legislature. The most likely date for the election is March 22nd, 2025. I suggest that we use March 21st as valid through and valid from date for now, and change it if the election happens on a different date. What do you think?

Sounds like a plan.

azuser (approved):

jpmckinney (approved):
Checked structure and basic facts (going from 39 to 43 constituencies).

PR #394, opened 2024-11-12
update DE constituencies for 2025 snap election

zikowang:
Adding README
- [with the list of constituencies for the upcoming german snap election in 2025](http://www.bundeswahlleiterin.de/dam/jcr/159ea2a8-6e4b-45a2-9726-af19163ef19f/bundeswahlgesetz_anlage2.pdf)
- [with links / documents for the changes](https://www.bundeswahlleiterin.de/bundestagswahlen/2025/wahlkreiseinteilung.html#c4a81b5d-3fd6-4062-a21e-c5835218c7a9)

**Why we propose to change the ocd_paths back using names instead of numbers**

1. while updating the ocd_paths we discovered that 184 constituency numbers changed while they actually staying the same. It was triggered by one change.
2. also we have 6 valid through and 6 valid from constituencies that keeps the number but is actually different (with a different name).
3. to keep backward compatibility we added entries to the `aliases.csv` file

**We can not build the new country CSV file because the alias file has chains**

Example:
```
sameAs chain: ocd-division/country:de/land:bb/wahlkreis:brandenburg_an_der_havel_~_potsdam-mittelmark_i_~_havelland_iii_~_teltow-fläming_i -> ocd-division/country:de/ed:60 -> ocd-division/country:de/ed:brandenburg_an_der_havel_-_potsdam-mittelmark_i_-_havelland_iii_-_teltow-fläming_i 
```
- the first two entries were added in 2021
- with our proposal, we would add another alias for `ocd-division/country:de/ed:60` using the name instead of the number

How can we proceed from here, since the election is near and we need those OCDs to official.

Best,


zikowang:
Thank you for the reply,

When we update the sameAs entry for `ocd-division/country:de/land:bb/wahlkreis:brandenburg_an_der_havel_~_potsdam-mittelmark_i_~_havelland_iii_~_teltow-fläming_i`

we would lose the `ocd-division/country:de/ed:60` in between, so that those paths would be invalid from now on.

Will that be a problem? If not, I am happy to adjust all those chain entries in the aliases file.

zikowang:
Okay, I cleaned up the aliases file and compiled the `country-de.csv` file.

Please have a look at the results.

identifiers/country-de/wahlkreise.csv
```diff
@@ -1,300 +1,306 @@
...
+ocd-division/country:de/ed:potsdam_-_potsdam-mittelmark_ii_-_teltow-fläming_ii,Potsdam - Potsdam-Mittelmark II - Teltow-Fläming II,,
+ocd-division/country:de/ed:62,Dahme-Spreewald - Teltow-Fläming III - Oberspreewald-Lausitz I,,2025-02-23
+ocd-division/country:de/ed:dahme-spreewald_-_teltow-fläming_iii,Dahme-Spreewald - Teltow-Fläming III,2025-02-23,
+ocd-division/country:de/ed:frankfurt_~oder~_-_oder-spree,Frankfurt (Oder) - Oder-Spree,,
+ocd-division/country:de/ed:cottbus_-_spree-neiße,Cottbus - Spree-Neiße,,
+ocd-division/country:de/ed:65,Elbe-Elster - Oberspreewald-Lausitz II,,2025-02-23
+ocd-division/country:de/ed:elbe-elster_-_oberspreewald-lausitz,Elbe-Elster - Oberspreewald-Lausitz,2025-02-23,
+ocd-division/country:de/ed:66,Altmark,,2025-02-23
+ocd-division/country:de/ed:altmark_-_jerichower_land,Altmark - Jerichower Land,2025-02-23,
+ocd-division/country:de/ed:67,Börde - Jerichower Land,,2025-02-23
```
jloutsenhizer, line 71: It looks like some of these retain the number-based ID. Shouldn't they all be switched to names?

identifiers/country-de/aliases.csv
```diff
@@ -15,603 +15,603 @@ ocd-division/country:de/land:sl,ocd-division/country:de/state:sl,Added for backw
...
+ocd-division/country:de/state:sh/cd:4,ocd-division/country:de/ed:rendsburg-eckernförde,Rendsburg-Eckernförde,,
+ocd-division/country:de/state:sh/cd:5,ocd-division/country:de/ed:kiel,Kiel,,
+ocd-division/country:de/state:sh/cd:6,ocd-division/country:de/ed:plön_-_neumünster,Plön - Neumünster,,
+ocd-division/country:de/state:sh/cd:7,ocd-division/country:de/ed:pinneberg,Pinneberg,,
+ocd-division/country:de/state:sh/cd:8,ocd-division/country:de/ed:segeberg_-_stormarn-mitte,Segeberg - Stormarn-Mitte,,
+ocd-division/country:de/state:sh/cd:9,ocd-division/country:de/ed:ostholstein_-_stormarn-nord,Ostholstein - Stormarn-Nord,,
+ocd-division/country:de/state:sl/cd:296,ocd-division/country:de/ed:saarbrücken,Saarbrücken,,
+ocd-division/country:de/state:sl/cd:297,ocd-division/country:de/ed:saarlouis,Saarlouis,,
+ocd-division/country:de/state:sl/cd:298,ocd-division/country:de/ed:st._wendel,St. Wendel,,
+ocd-division/country:de/state:sl/cd:299,ocd-division/country:de/ed:homburg,Homburg,,
```
jloutsenhizer, line 584: We should _also_ have the mapping from the `ed:<number>` to `ed:<name>` in aliases. So for example we should have `ocd-division/country:de/ed:299` equivalent to `ocd-division/country:de/ed:homburg`.

Or are these left out because of the ID shifts that happened with the latest changes by the government?

jloutsenhizer (changes requested):
I haven't thoroughly looked through the changes but left some comments on initial issues I saw.

Overall, I wanted to confirm that it is the name that carries the identity of the district, and not the number. So even though the number of many districts changed, since the name remains the same it'd be accurate to say people running for re-election are running in the same district and, if they win, have a continuous tenure?

HKSenior:
@jpmckinney can you take a look for approval?

identifiers/country-de/wahlkreise.csv
jpmckinney: It looks like the numbered ones are only those with `validThrough`.

identifiers/country-de/aliases.csv
jpmckinney: +1

identifiers/country-de/wahlkreise.csv
jpmckinney: That said, it would make sense to have the name instead, for consistency.

identifiers/country-de/aliases.csv
zikowang: So adding more mappings, so would have 3 aliases for `ocd-division/country:de/ed:homburg`?

```
ocd-division/country:de/land:sl/wahlkreis:homburg,ocd-division/country:de/ed:homburg
ocd-division/country:de/state:sl/cd:299,ocd-division/country:de/ed:homburg
ocd-division/country:de/ed:299,ocd-division/country:de/ed:homburg
```

identifiers/country-de/wahlkreise.csv
jloutsenhizer: Yeah Ideally, I think the final set in this file would all be using names instead of some still using numbers.

identifiers/country-de/aliases.csv
jloutsenhizer: Correct, the idea being that data using any of the old formats can be translated to the latest OCD IDs by using the mappings from the aliases file

identifiers/country-de/wahlkreise.csv
zikowang: - updated the numeric based IDs to name based.
- updated aliases file for those changes

identifiers/country-de/aliases.csv
zikowang: - added alias entries for all numeric based IDs pointing to the name based ones

jloutsenhizer (approved):
Updates from last comments LGTM.

I spotchecked a few existing, added, and removed constituencies and didn't notice anything wrong.

PR #400, opened 2025-01-16
Modify EU ocd ids

HKSenior:
The current EU file implementation uses the top level (aka the highest level district to describe an area, for example `ocd-division/country:us` for the US and `ocd-division'/country:at` for Austria) as non-canonical ids to form the EU region. There are some disadvantages to this:

1. We no longer conform to a single file describing a single country/region.
2. An ID can have attributes within the eu file but nothing set within the origin country file (i.e. the top level Austria id within the eu file has an end date of 2023-12-01 but the Austria file has nothing set).

The proposed change consists of adding `region:eu` to all ids within the EU file. Therefore the current id `ocd-division/country:ie/eup:dublin` will be changed to `ocd-division/region:eu/country:ie/eup:dublin`. The spec for division identifiers can be found [here](https://open-civic-data.readthedocs.io/en/latest/proposals/0002.html).


jpmckinney:
The spec is clear: 

> Identifiers are in the format `ocd-division/country:<country_code>(/<type>:<type_id>)*`

So, the spec prohibits anything other than `country:` under `ocd-division`.

----

Separately, I am not clear what you mean, specifically, by the EU region. The European Union is not static, and adds (most recently Croatia) and removes (UK) members. I don't think it would be an advantage to alias all of the UK, etc. whenever such changes occur.

If you mean the continent of Europe, its boundary is [not universally defined](https://en.wikipedia.org/wiki/Geography_of_Europe), which leads to inconsistent descriptions of which countries are in Europe. I don't think we can come up with clear criteria for inclusion (seeing as no one else ever has).

> We no longer conform to a single file describing a single country/region.
> An ID can have attributes within the eu file but nothing set within the origin country file (i.e. the top level Austria id within the eu file has an end date of 2023-12-01 but the Austria file has nothing set).

We should delete country-eu.csv and instead have one file per country. Each country can use whatever attributes make sense.

TLDR: Split country-eu.csv into individual country files. No changes needed to identifiers.

HKSenior:
I like the idea of moving the ids from the eu file to their corresponding country files. But I do want to confirm a few things before I make the changes:

1. We will required an eu file with a single id to represent the eu as a whole (`ocd-division/region:eu`). I understand the spec is clear on the first type being `country:<country_code>`, however the eu doesn't satisfy this condition and `region` has been the preferred type to describe the eu within the repo.
2. If we move all ids to their respective files without changes we lose context of what countries are apart of the eu. Therefore  a minimal change would be needed. For example, `ocd-division/country:it` within the eu file should become `ocd-division/country:it/region:eu` when added to the it file. Similarly, `ocd-division/country:pl/eup:9` should become `ocd-division/country:pl/region:eu/eup:9`.

I understand the spec doesn't leave a lot of wiggle room for defining a supranational union such as the eu.

jpmckinney:
1. I don't know why/how `ocd-division/region:eu` was selected. `region` is a type used within countries. I would prefer a different type that is unique to [supranational political unions](https://en.wikipedia.org/wiki/Supranational_union), like `union`. At present, [the EU is the only union in the world](https://en.wikipedia.org/wiki/Supranational_union#Other_international_organisations_with_some_degree_of_integration).

2. The `eup` type means European Parliament. Is the presence of `eup` divisions not enough to indicate EU membership?

HKSenior:
Stepping back a little bit:

I understand the [spec](https://open-civic-data.readthedocs.io/en/latest/proposals/0002.html) denotes the following:
> Identifiers are in the format `ocd-division/country:<country_code>(/<type>:<type_id>)*`

But we can agree the current spec doesn't take into consideration a supranational union such as the EU. Which makes sticking to the standard of having `country:<country_code>` as the first type within all ids difficult to adhere to. For the instance of the EU I recommend we continue with a single file containing `ocd-division/region:eu`.

**Note**: I'm using `region` as a placeholder here. I'm open to changing this to `union` or another type that better describes the EU. I would like for us to agree there are two top level types that can exist within the repo, `country` and `x` (to describe the EU).

----

If we are going to move the top level (aka `ocd-division/country:it`) ids from the eu file back to their respective country files they will need to be modified. This is necessary not to lose context of what countries are within the EU. I propose we create non-canonical (aliases) to the top level ids already present within the respective country files. For example, for IT the `top_level.csv` file will change to this:

```
id,sameAs,name
ocd-division/country:it,,Italy
ocd-division/region:eu/member:it,ocd-division/country:it,"Italy, Member of the European Union"
```

**Note**: I'm using `region` as a placeholder here.

I choose to add `member` due to Italy being a member of the EU. Adding two top level types (`country` and `x`) would be a bit excessive and `member` best describes Italy's position within the EU. Another option is `part` currently being used within the GB directory to describe the parts of Great Britain. 

----

A similar change will occur for the European districts defined within the EU dir. For example, the Ireland European districts will be changed to the following (within the IE file):

```
id,sameAs,name,validFrom,validThrough
ocd-division/region:eu/member:ie/eup:dublin,,Dublin,,
ocd-division/country:ie/eup:dublin,ocd-division/region:eu/member:ie/eup:dublin,Dublin,,2025
```

The reason for the additional `region` here is to follow the general practice of scoping down (from the largest to smallest district). The largest top level in this case would be `x` to describe the eu, then the eu member (in this case IE) and finally the European district which in this case is Dublin.

**Note**: I'm primarily making these changes as non-canonical ids to give others time to phase the old ids out and in with the new ids by the end of the year.

jpmckinney:
On nomenclature: The EU describes its members as Member States, so `member` sounds fine to me.

If representing EU membership is a task that this project wants to take on (not sure where this demand is coming from?), then `ocd-division/REPLACEME:eu/member:ie` with an alias to `ocd-division/country:ie` is the minimum required to:

1. Indicate membership
2. Associate the member to the country

So far, that's OK with me.

On the `eup` IDs, we already have e.g. `ocd-division/country:ie/eup:dublin`. I don't see why we need the alias `ocd-division/REPLACEME:eu/member:ie/eup:dublin`. If a program (a human wouldn't need this sort of assistance) starts with `ocd-division/REPLACEME:eu`, and descends to the child `ocd-division/REPLACEME:eu/member:ie`, it should follow the alias to `ocd-division/country:ie`, before continuing the descent, at which point it encounters the `ocd-division/country:ie/eup:dublin`, etc. IDs.

The way this project is structured, a program must have logic related to specific types. For example, `ocd-division/country:us/state:al/cd:10` is a congressional district. The person holding the position related to that division is a member of Congress, which is a US entity (`ocd-division/country:us`), NOT an Alabama entity (`ocd-division/country:us/state:al`) (as the ID might suggest). The spec (from scanning it quickly) gives no advice on how to arrange these parent-child relationships. For example, the US could have done `ocd-division/country:us/cd:al-10`, instead. Pro: It makes it clear that the Congressional district relates to a US political entity. Con: It requires parsing the type ID to determine the state. Another complication: Since the US is a federation, there is some sense in having the district "belong" to the state, since if the state secedes, it will take its district with it (it won't still have representation in Congress).

So, I think there is no need for `ocd-division/REPLACEME:eu/member:ie/eup:dublin`. A program will need to know what `eup` means (just like it needs to know what `cd` means), and once it knows that, it doesn't need anything more than the existing `ocd-division/country:ie/eup:dublin` ID.

HKSenior:
Updated changes to PR.

1. I also removed the uk id from the EU countries.csv file as Brexit happened back in 2020, [here](https://en.wikipedia.org/wiki/Brexit).
2. Updated the compile script to not throw an error when it encounters the top level (`ocd-division/region:eu`) as a missing parent due to the id being defined within the EU directory.

jpmckinney (approved):
LGTM

HKSenior:
Thanks for reviewing the PR @jpmckinney

PR #401, opened 2025-01-16
Update /identifiers/country-ca/aliases.csv to include -2023 numeric id to name based id aliases

evannjw:
Adding aliases between the numeric identifiers and the name based identifiers for the new 2023 districts. Ideally we'd use scripts/country-ca/ca_federal_electoral_districts.py to generate an updated set of current name based ids, but it looks like https://lop.parl.ca/sites/ParlInfo/default/en_CA/ElectionsRidings/Ridings won't have the new set of districts (active districts) until the next general election happens. I'm adding aliases for the districts that aren't being renamed, and will update this list with the renamed districts once those can be created with the script after the general election.


evannjw:
@jpmckinney Can you take a look?

jpmckinney (approved):
I have not checked for accuracy, but looks good structurally.

jloutsenhizer (approved):
Spot checked about a dozen and didn't find any errors

PR #408, opened 2025-05-08
adding state court and territory ocdids

tatekamish:


crweber-goog (commented):
For what it's worth, these changes look good to me. The addition of the header for `us_state_courts.csv` is good for consistency but a challenging diff for GitHub =).

My understanding of the validFrom/validThrough is that since these are net adds of existing courts/territories they're not needed, but using "Florida's Sixth District Court of Appeal" as an example, if this change was submitted prior to it going into operation (i.e. before 01 Jan 2023) then we'd want to have set the validFrom?

tatekamish:
Thanks for the feedback. I was unable to run `compile.py` without those added headers in `us_state_courts.csv`.

My understanding is yes, if we were to add the ID before the district was adopted, we would want to include the validFrom date. But since we are adding it after the fact, I don't think it's needed.

identifiers/country-us/us_state_courts.csv
```diff
@@ -1,1642 +1,1682 @@
...
+ocd-division/country:us/state:co/district_court:13,CO State District Court - 13th District,,,
+ocd-division/country:us/state:co/district_court:14,CO State District Court - 14th District,,,
+ocd-division/country:us/state:co/district_court:15,CO State District Court - 15th District,,,
+ocd-division/country:us/state:co/district_court:16,CO State District Court - 16th District,,,
+ocd-division/country:us/state:co/district_court:17,CO State District Court - 17th District,,,
+ocd-division/country:us/state:co/district_court:18,CO State District Court - 18th District,,,
+ocd-division/country:us/state:co/district_court:19,CO State District Court - 19th District,,,
+ocd-division/country:us/state:co/district_court:20,CO State District Court - 20th District,,,
+ocd-division/country:us/state:co/district_court:21,CO State District Court - 21st District,,,
+ocd-division/country:us/state:fl/district_court:6,Florida Appellate Court District 6,,,
```
jloutsenhizer, line 131: Could you either update this name or the other names for this district or the other FL district courts to be consistent?

identifiers/country-us/us_state_courts.csv
```diff
@@ -1,1642 +1,1682 @@
...
+ocd-division/country:us/state:ky/district_court:52,KY State District Court - 52nd District,,,
+ocd-division/country:us/state:ky/district_court:53,KY State District Court - 53rd District,,,
+ocd-division/country:us/state:ky/district_court:54,KY State District Court - 54th District,,,
+ocd-division/country:us/state:ky/district_court:55,KY State District Court - 55th District,,,
+ocd-division/country:us/state:ky/district_court:56,KY State District Court - 56th District,,,
+ocd-division/country:us/state:ky/district_court:57,KY State District Court - 57th District,,,
+ocd-division/country:us/state:ky/district_court:58,KY State District Court - 58th District,,,
+ocd-division/country:us/state:ky/district_court:59,KY State District Court - 59th District,,,
+ocd-division/country:us/state:ky/district_court:60,KY State District Court - 60th District,,,
+ocd-division/country:us/state:la/court_of_appeals:1,LA State Court of Appeals - 1st District,,,
```
jloutsenhizer, line 368: It sounds like these LA court of appeals district names were incorrect and these should have been labeled as circuits right? If so could you update these

identifiers/country-us/us_state_courts.csv
```diff
@@ -1,1642 +1,1682 @@
...
+ocd-division/country:us/state:la/court_of_appeals:2,LA State Court of Appeals - 2nd District,,,
+ocd-division/country:us/state:la/court_of_appeals:3,LA State Court of Appeals - 3rd District,,,
+ocd-division/country:us/state:la/court_of_appeals:4,LA State Court of Appeals - 4th District,,,
+ocd-division/country:us/state:la/court_of_appeals:5,LA State Court of Appeals - 5th District,,,
+ocd-division/country:us/state:la/court_of_appeals:1/district:1,"Louisiana Appellate Court Circuit 1, District 1",,
```
jloutsenhizer, line 373: We usually have the nesting of IDs based on the level of government that is determining the districts. My understanding is that the Lousiana state legislature sets the districts for each circuit. SO I think these should be be direct children of `ocd-division/country:us/state:la`. Perhaps something like:
`ocd-division/country:us/state:la/court_of_appeals:1_1` for circuit 1 district 1

jloutsenhizer (changes requested):
A few small requests, but overall looks good to me

identifiers/country-us/us_state_courts.csv
tatekamish: I think they are "Circuit Courts of Appeal" ex. https://www.la3circuit.org/

tatekamish:
Hi, I'm just following up on this since it says there is a workflow waiting approval.

identifiers/country-us/us_state_courts.csv
jpmckinney: I'm blocked from the official website by Cloudflare for whatever reason.

Since it's just the name column, I'm not too fussed with whichever name is chosen.

It's too bad the type has the typo "court_of_appeals" instead of "court_of_appeal" but it's not worth changing all the old OCDIDs.

jloutsenhizer (approved):

PR #411, opened 2025-06-19
Update EG districts for 2025

zikowang:
[Law No. 85 of 2025](http://www.alamiria.com/Sec/TashTxt?id=raGPzNh0Pio=)
Amending Some Provisions of the House of Representatives Law. It takes effect the day after its publication in the Official Gazette, which is on 2025-06-10 [ref](https://www.almasdar.com/156868).


zikowang:
@jpmckinney I updated the EG Readme file since [the almasdar source](https://www.almasdar.com/156868) has the original text as well (under the article itself)

@jloutsenhizer thanks for the review. File updated.

PR #385, opened 2024-05-31
UK - Constituency changes for next HoC general election

sguenther85:
After the 2023 Periodic Review of Westminster constituencies the  Parliamentary Constituencies Order 2023 the UK Gov introduced a lot of changes for the next general HoC election.

Including:
211 newly named constituencies (lost of redistricting)
Many abolished constituency names
Some disappearing and newly created seats

See Wiki for Summary and link below for law with table that lists new constituencies (pdf).
https://en.wikipedia.org/wiki/2023_Periodic_Review_of_Westminster_constituencies#New_and_abolished_constituencies
https://www.legislation.gov.uk/uksi/2023/1230/pdfs/uksi_20231230_en.pdf

More background:
https://commonslibrary.parliament.uk/constituency-boundary-review-data-for-new-constituencies/

Readme will follow. Due to the recent unexpected election (by date), i wanted to share the constituency changes first

PR #433, opened 2026-05-03
Add OCD-IDs for UK Scottish Parliament, Senedd, wards, and CEDs

MiniCodeMonkey:
## Summary

Closes the OCD-ID coverage gap that downstream consumers run into when working with UK political districts. `country-gb.csv` previously only carried Westminster (`/ed:`) constituencies; this PR adds the four largest missing categories:

| Type | Path | Rows |
|---|---|---|
| Scottish Parliament constituencies (`spc`) | `country:gb/part:sct/spc:<slug>` | 73 |
| Scottish Parliament regions (`spr`) | `country:gb/part:sct/spr:<slug>` | 8 |
| Senedd / Welsh Parliament constituencies (`senc`) | `country:gb/part:wls/senc:<slug>` | 40 (current) + 16 (May 2026) |
| Senedd electoral regions (`senr`) | `country:gb/part:wls/senr:<slug>` | 5 (abolished May 2026) |
| Electoral wards (`ward`) | `country:gb/part:<nation>/ward:<gss>` | 8,396 UK-wide |
| County electoral divisions (`ced`) | `country:gb/part:eng/ced:<gss>` | 1,366 (England only) |

Total: ~9,900 new identifiers. NI Assembly is intentionally not touched -- post the 2023 OSNI review the boundaries are byte-identical to NI Westminster constituencies, so those `ed:` rows already serve.

## Conventions

The path scheme follows the existing merged convention in `constituencies.csv` (extended most recently in #385). I've also added a `gss_code` extra column on every row that maps to a single ONS GSS code, mirroring the older `country-uk` files and #186 so consumers can crosswalk OCD-IDs ↔ GSS codes without a separate lookup.

A few specific calls worth flagging for review:

* **Type prefixes follow ONS column codes**: `spc` / `spr` (Scottish Parliament constituency / region), `senc` / `senr` (Senedd constituency / region), `ward`, `ced`. These match modern ONS naming -- in particular `senc` / `senr` rather than the older `nawc` / `nawr` because the body was renamed from "National Assembly for Wales" to "Senedd / Welsh Parliament" in 2020 and ONS now ships `senc22cd` / `senr22cd` columns. Happy to switch to `nawc` / `nawr` for backwards parity with #186 if that's preferred.

* **Hierarchy retained** (`country:gb/part:sct/spc:slug`) for consistency with the current `constituencies.csv` rather than going flat (`country:gb/spc:slug`) per #186. The original objection in #184 that hierarchy implies geographic scope is specifically about Westminster (a UK-wide body); for Scottish Parliament / Senedd the part-nation nesting is geographically truthful. Either way the typed prefix carries the disambiguation, so I'm fine flipping to flat if reviewers want to pull the trigger on the #186 direction in this PR -- just let me know.

* **Wards keyed on GSS, not name**, because UK ward names collide nationally (21 distinct "Castle" wards). They're nested under `part:<nation>` rather than `lad:` or `council:` since this repo doesn't yet have UK Local Authority District identifiers; adding those is a much bigger change and out of scope here.

* **Senedd 2026 transition** marked with `validThrough=2026-05-05` on the 40+5 old rows and `validFrom=2026-05-06` on the new 16 rows. The new rows use placeholder GSS codes `W92000001`-`W92000016` (DataMapWales' internal numbering) because ONS hasn't published canonical 2026 GSS codes yet; once they do, those rows will need a `sameAs` plus a refreshed `gss_code`.

## Sources

* Westminster, Scottish Parliament, Senedd (current), wards, CEDs: ONS [Open Geography Portal](https://geoportal.statistics.gov.uk/) -- Open Government Licence v3.0.
* Senedd post-May-2026 constituencies: [DataMapWales](https://datamap.gov.wales/layers/geonode:senedd_final_2026) (Welsh Government, OGL v3.0).

## Validation

* `python3 test/compile_test.py` passes (TEST_TMPDIR set).
* Recompiled `identifiers/country-gb.csv` is included in the diff.
* All new IDs validate against the regex in `scripts/compile.py`.

## Refs

* #170 -- best-practices discussion
* #184 -- country code + hierarchy debate
* #186 -- chris48s' stalled flat-scheme migration; this PR adopts the same `gss_code` column convention.
* #385 -- 2024 Westminster boundary changes (most recent merged precedent for the hierarchy used here).

cc @chris48s @jpmckinney @sguenther85 -- given the unresolved conventions discussion in #170 / #184, would appreciate your steer on the hierarchy + prefix choices before this gets too far.


jpmckinney:
#184 remains unresolved, but that's about what to do about the old uk.

In the meantime, I'd say we are unconstrained about gb.

Per the discussion in #184, I don't see a principled reason to have a proliferation of types like spc, spr, senc, senr, etc. especially when these types get renamed whenever a political body chooses to rename itself.

My understanding is that ONS organizes itself such that anything with the label "senc" is a Senedd / Welsh Parliament constituency – such that if the maintainers for Canada used that type for a different purpose, it would imply that there is a Senedd / Welsh Parliament constituency in Canada. This project doesn't organize itself that way.

As such, we can simplify and use e.g. `ed` (electoral district) for constituency (like in most countries), `region` for regions, and `ward` for wards.

GSS sounds fine to me. Canada also used numeric codes, but when these became unstable across vintages (and when we wanted to extend coverage into the past, for which we had no IDs), we switched to full names (and aliased the old numbers). It sounds like for gb we'd have to add local authorities to avoid numbers.

